Issue #3, opened 2017-02-12
Review UploadedFile and ServerRequest


Nyholm:
See if we can remove features and add type hints

Zegnat:
Now that there are factories available, would it be acceptable to clean-up ServerRequest by moving `fromGlobals` and `getUriFromGlobals` to ServerRequestFactory and UriFactory respectively? This means ServerRequest can only be instantiated by `__construct` or through the factory, which makes sense to me.

If moving `getUriFromGlobals` to UriFactory is acceptable I would propose a slightly different signature:

```php
public function createUriFromArray(array $server)
```

This would mirror ServerRequestFactory’s `createServerRequestFromArray` and gets rid of the hard dependency on a global `$_SERVER`. That in turn would allow us to get rid of the `$SERVER = $_SERVER`, `$_SERVER = $server` jumble.

This only leaves `normalizeFiles` on the ServerRequest as a bit of an odd-one out, but I haven’t really looked into UploadedFile yet.

Let me know if a PR for the above would be welcome.

Nyholm:
Great. Lets move away from $_GLOBALS. 
I would be happy to review a PR.

PR #5, opened 2017-04-10
Move instantiation code to factories.

Zegnat:
**WORK IN PROGRESS, DO NOT MERGE.**

First and foremost: this changes the public API for ServerRequest, BC breaks need to be documented. (Or not, seeing how we are at a version below 1.0.0.)

This implements the changes from https://github.com/Nyholm/psr7/issues/3#issuecomment-292812159 but should be discussed further before merge.

1. `getUriFromGlobals` has been renamed `createUriFromArray` and has been moved from ServerRequest to UriFactory.
   
   The signature has been changed to accept a single `$_SERVER`-structure as argument, instead of the method depending directly on the global `$_SERVER` variable.
2. `fromGlobals` has been renamed `createServerRequestFromGlobals` and has been moved from ServerRequest to ServerRequestFactory.
   
   The signature has been changed to accept 5 separate array structures matching the different global variables.
3. `normalizeFiles`, `createUploadedFileFromSpec`, and `normalizeNestedFileSpec` have been moved from ServerRequest to ServerRequestFactory. This is probably not the perfect place for them either, but gets them out of ServerRequest.
4. Return type hinting has been added to the factories.

`createUriFromArray` sets no default scheme value, but [http-interop/http-factory-tests](https://github.com/http-interop/http-factory-tests) expects ServerRequests to default to `http` through the factory. I made `createServerRequestFromGlobals` follow the factory expectation and added the default scheme there as well. This is a change from the old `fromGlobals`. But I am not sure if it is a good idea or not. Is there a use-case for creating a ServerRequest without a scheme?

The factory interface’s `createServerRequestFromArray` and our `createServerRequestFromGlobals` differ also on their requirement for a method. PSR-17 has us throw an InvalidArgumentException when the method can’t be found in the array, while `createServerRequestFromGlobals` defaults to `GET`. Should we follow the factory expectation here as well?

If we are following the factory expectation on both counts, it might be best to have `createServerRequestFromGlobals` go through `createServerRequestFromArray` first. Then adding the other globals to the created ServerRequest.

What would people expect to happen?

Finally: why is `normalizeFiles` public? Do we mean to support its API during future updates? If not, it should really be marked private.

### DIFF
--- a/src/Factory/ServerRequestFactory.php
+++ b/src/Factory/ServerRequestFactory.php
@@ -25,29 +25,21 @@ class ServerRequestFactory implements ServerRequestFactoryInterface
 
     public function createServerRequestFromArray(array $server): ServerRequestInterface
     {
-        if (!isset($server['REQUEST_METHOD'])) {
-            throw new InvalidArgumentException('Cannot determine HTTP method');
-        }
-        $method = $server['REQUEST_METHOD'];
-
-        $uri = (new UriFactory())->createUriFromArray($server);
-        if ($uri->getScheme() === '') {
-            $uri = $uri->withScheme('http');
-        }
-
-        $protocol = isset($server['SERVER_PROTOCOL']) ? str_replace('HTTP/', '', $server['SERVER_PROTOCOL']) : '1.1';
-
-        return new ServerRequest($method, $uri, [], null, $protocol, $server);
+        return new ServerRequest(
+            $this->getMethodFromEnvironment($server),
+            $this->getUriFromEnvironmentWithHTTP($server)
+        );
     }
 
     /**
      * Create a new server request from a set of arrays.
      *
-     * @param array $server Typically $_SERVER or similar structure.
-     * @param array $cookie Typically $_COOKIE or similar structure.
-     * @param array $get    Typically $_GET or similar structure.
-     * @param array $post   Typically $_POST or similar structure.
-     * @param array $files  Typically $_FILES or similar structure.
+     * @param array $server  Typically $_SERVER or similar structure.
+     * @param array $headers Typically the output of getallheaders() or similar structure.
+     * @param array $cookie  Typically $_COOKIE or similar structure.
+     * @param array $get     Typically $_GET or similar structure.
+     * @param array $post    Typically $_POST or similar structure.
+     * @param array $files   Typically $_FILES or similar structure.
      *
      * @throws InvalidArgumentException If no valid method or URI can be determined.
      *
@@ -55,24 +47,16 @@ class ServerRequestFactory implements ServerRequestFactoryInterface
      */
     public function createServerRequestFromArrays(
         array $server,
+        array $headers,
         array $cookie,
         array $get,
         array $post,
         array $files
     ): ServerRequestInterface {
-        if (!isset($server['REQUEST_METHOD'])) {
-            throw new InvalidArgumentException('Cannot determine HTTP method');
-        }
-        $method = $server['REQUEST_METHOD'];
-
-        $uri = (new UriFactory())->createUriFromArray($server);
-        if ($uri->getScheme() === '') {
-            $uri = $uri->withScheme('http');
-        }
+        $method = $this->getMethodFromEnvironment($server);
+        $uri = $this->getUriFromEnvironmentWithHTTP($server);
 
         $protocol = isset($server['SERVER_PROTOCOL']) ? str_replace('HTTP/', '', $server['SERVER_PROTOCOL']) : '1.1';
-
-        $headers = function_exists('getallheaders') ? getallheaders() : [];
 
         $serverRequest = new ServerRequest($method, $uri, $headers, null, $protocol, $server);
 
@@ -85,6 +69,10 @@ class ServerRequestFactory implements ServerRequestFactoryInterface
 
     /**
      * Create a new server request from the current environment variables.
+     * Defaults to a GET request to minimise the risk of an InvalidArgumentException.
+     * Includes the current request headers as supplied by the server through `getallheaders()`.
+     *
+     * @throws InvalidArgumentException If no valid method or URI can be determined.
      *
      * @return ServerRequestInterface
      */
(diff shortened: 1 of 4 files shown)


src/Factory/ServerRequestFactory.php
```diff
@@ -4,35 +4,154 @@
...
+use Nyholm\Psr7\UploadedFile;
 
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  * @author Martijn van der Ven <martijn@vanderven.se>
  */
 class ServerRequestFactory implements ServerRequestFactoryInterface
 {
-    public function createServerRequest($method, $uri)
+    public function createServerRequest($method, $uri): ServerRequestInterface
```
Nyholm: Good, lets do PHP7 syntax where ever we can.

Zegnat:
> I believe that the only use case for createServerRequestFromGlobals is in your front end controller. I think we should make it as easy as possible. Do you think it is okey to use Globals in this factory method only?

Maybe it is a personal bias, but I am not a fan of having methods that solely rely on variables in an outer scope. Interestingly, Symfony doesn’t do a lot more than collect the globals and [put them through to their actual factory](https://github.com/symfony/symfony/blob/v3.2.7/src/Symfony/Component/HttpFoundation/Request.php#L290):

```php
public static function createFromGlobals()
{
    // …
    $request = self::createRequestFromFactory($_GET, $_POST, array(), $_COOKIE, $_FILES, $server);
    // …
}
```

Maybe that is the real trick here? Have a `createServerRequestFromArrays` and `createServerRequestFromGlobals`, where the second simply aliases the first with prefilled arguments? Another option is to make the arguments optional on `createServerRequestFromGlobals` and have them default to their superglobals when `null`.

The latter would allow users to do `(new ServerRequestFactory())->createServerRequestFromGlobals()` while tests can be written using `(new ServerRequestFactory())->createServerRequestFromGlobals($server, $cookies /*, etc. */)`.

Nyholm:
> Maybe that is the real trick here? Have a createServerRequestFromArrays and createServerRequestFromGlobals, where the second simply aliases the first with prefilled arguments? 

I like that!

> The latter would allow users to do (new ServerRequestFactory())->createServerRequestFromGlobals() while tests can be written using (new ServerRequestFactory())->createServerRequestFromGlobals($server, $cookies /*, etc. */).

But if `createServerRequestFromGlobals` is just an alias for `createServerRequestFromArrays` once could just use `createServerRequestFromArrays` for testing, right?

Zegnat:
> But if `createServerRequestFromGlobals` is just an alias for `createServerRequestFromArrays` once could just use `createServerRequestFromArrays` for testing, right?

Yes. The only difference between the two options is that one uses two methods and no optional arguments, while the other combines it into a single method with optional arguments. I prefer the first option personally. I’ll give implementation a go!

Nyholm:
Awesome, please do!

src/Factory/ServerRequestFactory.php
```diff
@@ -4,35 +4,178 @@
...
+        $method = $server['REQUEST_METHOD'];
+
+        $uri = (new UriFactory())->createUriFromArray($server);
+        if ($uri->getScheme() === '') {
+            $uri = $uri->withScheme('http');
+        }
+
+        $protocol = isset($server['SERVER_PROTOCOL']) ? str_replace('HTTP/', '', $server['SERVER_PROTOCOL']) : '1.1';
+
+        $headers = function_exists('getallheaders') ? getallheaders() : [];
```
Zegnat: I am wondering if we should take $headers as another argument of `createServerRequestFromArrays` and only fill it with `getallheaders` inside `createServerRequestFromGlobals`. That way, no magic is going on in `createServerRequestFromArrays` and it is truly an extension on `createServerRequestFromArray`.

src/Factory/ServerRequestFactory.php
```diff
@@ -4,35 +4,178 @@
...
+    /**
+     * Create a new server request from the current environment variables.
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequestFromGlobals(): ServerRequestInterface
+    {
+        $server = $_SERVER;
+        if (false === isset($server['REQUEST_METHOD'])) {
+            $server['REQUEST_METHOD'] = 'GET';
```
Zegnat: For simplicity of `(new ServerRequestFactory())->createServerRequestFromGlobals()`, we make sure there is a method available. If users want to use `createServerRequestFromArrays` directly, they will have to make sure of it themselves.

src/Factory/ServerRequestFactory.php
```diff
@@ -4,35 +4,178 @@
...
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequestFromArrays(
+        array $server,
+        array $cookie,
+        array $get,
+        array $post,
+        array $files
+    ): ServerRequestInterface {
```
Zegnat: The first 12 lines of `createServerRequestFromArrays` are an exact copy of `createServerRequestFromArray`, just so they are equalised. What would be the best way to reduce this duplicate code?

I first thought I would do `$this->createServerRequestFromArray($server)` on the first line, but that makes adding the headers harder.

Zegnat:
Hmm, headers keep bugging me a little.

All headers are contained within `$_SERVER`. But I can’t seem to find if PSR-17’s `createServerRequestFromArray` expects them to be extracted them from the passed array. So the question is, should it have headers?

```
$serverRequest = (new ServerRequestFactory())->createServerRequestFromArray($_SERVER);
count($request->getHeaders()); // What will this be?
```

And if it expects headers, how will we set them? As per `testCreateServerRequestDoesNotReadServerSuperglobal` it is specifically specified that `$_SERVER` cannot be used and `createServerRequestFromArray` needs to be limited to the array that is passed to it. I would say this means `getallheaders` cannot be used, as it does not consider the passed array and goes straight to the server request. This probably means `getallheaders` can only be used for our own `createServerRequestFromGlobals`.

Any ideas?

Zegnat:
Some weird moving of the commits seem to have happened, but what ever.

@Nyholm, please review the code as it stands now. The cleanest I have been able to come up with. It also best conforms to what PSR-17 seems to want of `createServerRequestFromArray`.

src/Factory/ServerRequestFactory.php
```diff
@@ -4,35 +4,184 @@
...
+     * @param array $cookie  Typically $_COOKIE or similar structure.
+     * @param array $get     Typically $_GET or similar structure.
+     * @param array $post    Typically $_POST or similar structure.
+     * @param array $files   Typically $_FILES or similar structure.
+     *
+     * @throws InvalidArgumentException If no valid method or URI can be determined.
+     *
+     * @return ServerRequestInterface
+     */
+    public function createServerRequestFromArrays(
```
Nyholm: Hm, how do you feel about this name?

Maybe change it to something more expressive? Or maybe less expressive?

`createFromMultipleArrays`?

Zegnat: > Hm, how do you feel about this name?

I have no strong feelings about it. I was merely following the factory’s `createServerRequestFromArray`. PSR-17 has a good rationale for including “ServerRequest” in the method names, though that doesn’t apply to us.

I can change it to anything you’d like, really.

Zegnat:
> Let me know when you are ready for a merge.

Ready when you are. I would rather get this into the repo and iterate on it later when new ideas come up, than freeze it here until everything is absolutely perfect. Especially for < 1.0.0 versions.

Zegnat:
Side-note: it is hard to strike the balance between a pure PSR-7 / PSR-17 implementation and usability of the code as a library. A completely separate way of solving this is to not implement anything on the “user friendly” side and offer different classes that do that. E.g. [Zegnat\Utils\ ServerRequestFromGlobals](https://gist.github.com/Zegnat/6fd5c1774be4a597772b4c8febb0958a).

Nyholm:
Hm, Im also not sure. I'll merge this for now and we can see how it feels. 

Thank you!

PR #9, opened 2017-04-17
Update composer.json with PSR-17 information

Zegnat:
Mostly small changes to make it easier for people to find this package.

Start providing `psr/http-factory-implementation`, which seems to be the virtual package most other implementers are using: [`psr/http-factory-implementation` on Packagist](https://packagist.org/providers/psr/http-factory-implementation).


Nyholm:
Awesome. Thank you.

PR #26, opened 2018-02-06
Allow string values to FilterPort

Nyholm:


Zegnat:
Is there a reason for removing typing again?

Nyholm:
Yeah, All the tests failed =)

It was a mistake by me merging #24 this morning. 

The PSR specifies that a \InvalidArgument should be thrown. We generated a lot of TypeErrors.

Nyholm:
Btw, do you have any other suggestion?

Zegnat:
> The PSR specifies that a \InvalidArgument should be thrown. We generated a lot of TypeErrors.

Ugh, I keep forgetting we can’t just use PHP 7+ type hinting while also following the Interface and its description 🤦 Carry on. Haha.

PR #27, opened 2018-02-06
Added changelog for 0.3.0

Nyholm:
We should release 0.3.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,22 @@
+# Changelog
+
+All notable changes to this project will be documented in this file, in reverse chronological order by release.
+
+## 0.3.0
+
+### Added
+
+- Return types.
+- Many `InvalidArgumentException`s are thrown when you use invalid arguments. 
+- Integration tests for `UploadedFile` and `ServerRequest`.
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+### Added
+
+- Return types.
+- Many `InvalidArgumentException`s are thrown when you use invalid arguments. 
+- Integration tests for `UploadedFile` and `ServerRequest`.
+
+### Changed
+
+- We dropped PHP7.0 support. 
+- PSR-17 factories has been marked as internal. They do not fall under or BC promise until PSR-17 is accepted.
```
Zegnat: Should be “under _our_ BC”.
(Is it “factories has been” or “factories have been”? That always gets me.)

Zegnat:
All the changes looked good to me, so no complaints on making them accessible as a new release. I’m in the release and iterate often camp anyway, and we can always make patch releases if we bump into anything.

Nyholm:
Thank you for the review

CHANGELOG.md
Nyholm: Thank you

PR #28, opened 2018-03-07
Require php-http/discovery

DavidGarciaCat:
This update should fix https://github.com/mailgun/mailgun-php/issues/452#issuecomment-371197710

Then we might need to release a new version and upgrade Mailgun's version of `Nyholm/psr7`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
     "require": {
         "php": "^7.1",
         "psr/http-message": "^1.0",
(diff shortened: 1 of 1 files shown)


NickvdMeij:
This will **not** fix https://github.com/mailgun/mailgun-php/issues/452.

Nyholm:
This library does not have a dependency on php-http/discovery. Im following up the discussion in php-http/discovery#106

PR #29, opened 2018-03-17
Removed Discovery strategy since it is included in php-http/discovery

Nyholm:


Zegnat:
Looks good to me, commits that *remove* code from a package meant to be a minimal viable implementation are always welcome :wink:

Nyholm:
Thanks for the review.

PR #32, opened 2018-03-29
Make all classes final

### LINKED ISSUE #31: Make message classes final
From https://github.com/guzzle/psr7/pull/158#issuecomment-376237838

> I'm pretty confident that I will propose making all message classes in this project final if we ever hit a 2.0 version.
> 
> Reason: HTTP messages are represented as value objects, implementing an interface. From this point I don't see any valid use cases for extending the message classes.

FYI @sagikazarmark

Nyholm:
This will fix #31. 

Im not sure I like the "trick" with the `RequestTrait` I had to do because the `ServerRequest` was extending `Request. `

### DIFF
--- a/src/ServerRequest.php
+++ b/src/ServerRequest.php
@@ -13,8 +13,11 @@ use Psr\Http\Message\UriInterface;
  * @author Michael Dowling and contributors to guzzlehttp/psr7
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class ServerRequest extends Request implements ServerRequestInterface
+final class ServerRequest implements ServerRequestInterface
 {
+    use MessageTrait;
+    use RequestTrait;
+
     /** @var array */
     private $attributes = [];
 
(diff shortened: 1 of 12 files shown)


Zegnat:
I am pro-`final`. But isn’t that fact that ServerRequest extends Request an example that there are valid extensions possible? So what’s wrong with Request not being final?

src/ServerRequest.php
```diff
@@ -34,11 +37,11 @@ class ServerRequest extends Request implements ServerRequestInterface
...
-     * @param string                               $method       HTTP method
-     * @param string|UriInterface                  $uri          URI
-     * @param array                                $headers      Request headers
-     * @param string|null|resource|StreamInterface $body         Request body
-     * @param string                               $version      Protocol version
+     * @param string                               $method  HTTP method
+     * @param string|UriInterface                  $uri     URI
+     * @param array                                $headers Request headers
+     * @param string|null|resource|StreamInterface $body    Request body
+     * @param string                               $version Protocol version
```
gmponos: I guess you either have to align them or not. (they are not aligned with $serverParams)

gmponos:
I guess this is a Breaking change right?

sagikazarmark:
I agree with @Zegnat . ServerRequest is actually a Request (as per the interface as well), so I don't see any harm in leaving it non-final.

(I understand though that you want to to avoid possible future complaints/questions asking why the rest of the classes are final...that's a pain in the @ ss, speaking of experience)

sagikazarmark:
Further thinking about this: I think it's enough that the inheritance appears in the contract. Noone will (should!) ever check that in the implementation. That being said, I see no harm in it either (except maybe that Request wouldn't be final anymore)

Nyholm:
I've actually revisited this PR many times without making a decision. I think Mark is correct in his final comment. People should not care about the implementation, they should rely on the interfaces. 

Specifically since this implementation is all about being strict PSR7 and no extra features. 

Thank you for the reviews.

PR #33, opened 2018-03-29
Added strict comparisons and some type annotations

Nyholm:
TODO: 
- [ ] Validate the type annotations with the specification.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -35,11 +35,11 @@ class Request implements RequestInterface
      * @param string                               $version Protocol version
      */
     public function __construct(
-        $method,
+        string $method,
         $uri,
         array $headers = [],
         $body = null,
-        $version = '1.1'
+        string $version = '1.1'
     ) {
         if (!($uri instanceof UriInterface)) {
             $uri = new Uri($uri);
@@ -66,10 +66,10 @@ class Request implements RequestInterface
         }
 
         $target = $this->uri->getPath();
-        if ('' == $target) {
+        if ('' === $target) {
             $target = '/';
         }
-        if ('' != $this->uri->getQuery()) {
+        if ('' !== $this->uri->getQuery()) {
             $target .= '?'.$this->uri->getQuery();
         }
 
@@ -130,7 +130,7 @@ class Request implements RequestInterface
     {
         $host = $this->uri->getHost();
 
-        if ('' == $host) {
+        if ('' === $host) {
             return;
         }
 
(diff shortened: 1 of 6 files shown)


src/Request.php
```diff
@@ -66,10 +66,10 @@ public function getRequestTarget(): string
         }
 
         $target = $this->uri->getPath();
-        if ('' == $target) {
+        if ('' === $target) {
             $target = '/';
         }
-        if ('' != $this->uri->getQuery()) {
+        if ('' !== $this->uri->getQuery()) {
```
sagikazarmark: Why Yoda in an expression like this? (Just asking, not looking for a flame war, though IMO this is too much consistency)

src/Uri.php
```diff
@@ -143,7 +143,7 @@ public function withScheme($scheme): self
     public function withUserInfo($user, $password = null): self
     {
         $info = $user;
-        if ('' != $password) {
+        if ('' !== $password) {
```
sagikazarmark: Password can be null

src/Response.php
```diff
@@ -137,7 +137,7 @@ public function withStatus($code, $reasonPhrase = ''): self
 
         $new = clone $this;
         $new->statusCode = (int) $code;
-        if ('' == $reasonPhrase && isset(self::$phrases[$new->statusCode])) {
+        if ('' === $reasonPhrase && isset(self::$phrases[$new->statusCode])) {
```
sagikazarmark: Nothing protects from passing a null reason phrase here which would bypass this condition

src/Uri.php
Zegnat: Good catch.

Something that I didn’t even think about: is empty string (`''`) a valid HTTP password? I.e. is `https://zegnat:@example.com/` different from `https://zegnat@example.com/`? If an empty string is valid, this if-clause should specifically check `null !== $password`.

sagikazarmark: I think it is valid, but I'm not sure it's any different from an no password at all. One could check the RFCs, but my gut tells that implementations might not always follow the RFC here.

Zegnat: A quick look at [RFC 3986](https://tools.ietf.org/html/rfc3986), as cited by PSR-7, tells me that an empty string indicates no password. So really, `''` equals `null` as far as URIs are concerned. This also matches with [the WHATWG URL Standard](https://url.spec.whatwg.org/) (that tries to bring several other standards together), where [password is an empty string by default](https://url.spec.whatwg.org/#concept-url-password).

Part of me wants to argue that the method signature is wrong, and should have been `withUserInfo($user, $password = '')`.

Nyholm:
Thank you for the reviews. I've made some updates.

PR #34, opened 2018-04-09
Support using numeric headers

iambrosi:
Support using numeric header values without having type errors. An example use case would be the usage of the `Content-Length` header, which might be set with the result of a call to `strlen`.

In fact, this is the case that caught my attention, by using the latest release of the [HWIOAuthBundle](https://github.com/hwi/HWIOAuthBundle). They [use `strlen` to set the `Content-Length` header](https://github.com/hwi/HWIOAuthBundle/blob/a45c54aed8acd2da032e7e03ae53f7e44d2a384e/OAuth/ResourceOwner/AbstractResourceOwner.php#L253) when building the request headers.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -211,7 +211,7 @@ trait MessageTrait
      */
     private function trimHeaderValues(array $values): array
     {
-        return array_map(function ($value) {
+        return array_map(function (string $value) {
             return trim($value, " \t");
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you. I like this. I thought the test would pass without this patch though... I will do some testing with your tea code without the patch before I merge.

PR #35, opened 2018-04-16
Remove check for empty header string value

sunkan:
I was using Buzz when I got an exception about empty header value.  
When I tried to use Zend Diactoros instead of this library it worked as expected.

If you'r interested I can put together a better fix that dose the validating more like it's done in [Zend Diactoros](https://github.com/zendframework/zend-diactoros/blob/master/src/HeaderSecurity.php#L96-L120)

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -85,7 +85,7 @@ trait MessageTrait
         }
 
         foreach ($value as $v) {
-            if (!is_string($v) || '' === trim($v)) {
+            if (!is_string($v)) {
                 throw new \InvalidArgumentException('Header values must be non-empty strings');
             }
         }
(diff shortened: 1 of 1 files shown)


Nyholm:
You are correct. Thank you for this fix.

Zegnat:
Should the Exception messages be updated after this, if this now allows empty header values? (Also interesting how this could be done without tripping up any tests, clearly something not being tested.)

Nyholm:
Exception messages should not be updated. The PR enforce that empty headers should NOT be added. 

About tests: 
Yes, I'm also curious. This PR adds checks for `$request->withAddedHeader('name', ['foobar', '']);`
I know we have tests for `$request->withAddedHeader('name', '');`

Nyholm:
Oh. I misread this PR. Martin, you are correct. We should update exception message.

Nyholm:
Sunkan, thank you for the chat yesterday and sorry for prematurely merging this PR. 

Could you improve the code by updating the exception messages and reference the RFC that specifies the header value? (Or where it says the value can be empty)

PR #38, opened 2018-04-18
Rephrase exception messages related to non empty header values

sunkan:
It looks like the [RFC-7230](https://tools.ietf.org/html/rfc7230#section-3.2) require header-value to be at least 1 charecter long. But I might be wrong.

What I can find most clients/librarys allow sending of empty header values

 - curl allow you to send empty headers
 - Firefox allows it for xhr/fetch with the motivation `We need it to be compliant with the XHR spec: "The empty string is legal and represents the empty header value."`
 - From the firefox thread it seams like Chrome also support sending and reciving empty header values
 - Guzzle support sending empty header-value https://github.com/guzzle/guzzle/pull/1915

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -77,16 +77,16 @@ trait MessageTrait
         if (!is_array($value)) {
             $value = [$value];
         } elseif (empty($value)) {
-            throw new \InvalidArgumentException('Header values must be non-empty strings');
+            throw new \InvalidArgumentException('Header values must be strings');
         }
 
         if (!is_string($header) || empty($header)) {
-            throw new \InvalidArgumentException('Header name must be non-empty strings');
+            throw new \InvalidArgumentException('Header name must be strings');
         }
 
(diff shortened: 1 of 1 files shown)


Zegnat:
Ha, that RFC feels a little misleading. It says:

> Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.

Which sure does sound like a “field value” is mandatory. But then they define field value as:

```
     field-value    = *( field-content / obs-fold )
```

Note the `*` there. A `field-value` contains 0 to infinite ([RFC 5234, §3.6.](https://tools.ietf.org/html/rfc5234#section-3.6)) other things. Zero. An empty `field-value` is a valid `field-value`.

So yes, we should definitely allow empty headers when following the RFC that PSR-7 it written against. Nice work! 👍

Nyholm:
Excellent. Thank you both!

PR #41, opened 2018-04-23
Use the local $server variable instead global $_SERVER

rancoud:
Fix a little mistake in **ServerRequestFactory** class about **createServerRequestFromGlobals()** function.


Nyholm:
You are very much correct. Thank you for finding this bug!

👍

PR #48, opened 2018-07-23
Better validate headers

### LINKED ISSUE #40: enforce header value as string ?
Hi,

Shouldn't we cast the header value as string in the trimHeaderValues() function ? 

A few http headers represent a number of seconds or bytes, and the user could easily does something like this : 
`new Response(200, ['Age' => 60]);`

instead of using a string for the value. In this case the trimHeaderValues() will crash trying to da a trim on a non string.

It seems more logical to cast the value as string before trying to do the trim, no ? or perhaps doesn't apply the trim if the value is not a string ? or just let it crash ?

what is you opinion ?

Nyholm:
This will fix #40

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -174,7 +174,7 @@ trait MessageTrait
         if (!is_array($values)) {
             $values = [$values];
         } elseif (empty($values)) {
-            throw new \InvalidArgumentException('Header values must be an array of strings, empty array given.');
+            throw new \InvalidArgumentException('Header values must be a string or an array of strings, empty array given.');
         } else {
             // Non empty array
             $values = array_values($values);
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -148,13 +148,20 @@ trait MessageTrait
     }
 
     /**
-     * Make sure header has a non-empty string name and a sting value.
+     * Make sure the header complies with RFC 7230.
      *
-     * Trims whitespace from the header values. Spaces and tabs ought to be
-     * excluded by parsers when extracting the field value from a header field.
+     * Header names must be a non-empty string consisting of token characters.
+     *
+     * Header values must be strings consisting of visible characters with all optional
+     * leading and trailing whitespace stripped. This method will always strip such
+     * optional whitespace. Note that the method does not allow folding whitespace within
+     * the values as this was deprecated for almost all instances by the RFC.
      *
      * header-field = field-name ":" OWS field-value OWS
+     * field-name   = 1*( "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^"
+     *              / "_" / "`" / "|" / "~" / %x30-39 / ( %x41-5A / %x61-7A ) )
      * OWS          = *( SP / HTAB )
+     * field-value  = *( ( %x21-7E / %x80-FF ) [ 1*( SP / HTAB ) ( %x21-7E / %x80-FF ) ] )
      *
      * @see https://tools.ietf.org/html/rfc7230#section-3.2.4
      */
@@ -163,21 +170,21 @@ trait MessageTrait
         if (!is_array($values)) {
             $values = [$values];
         } elseif (empty($values)) {
-            throw new \InvalidArgumentException('Header values must be strings, empty array given.');
+            throw new \InvalidArgumentException('Header values must be an array of strings, empty array given.');
         } else {
             // Non empty array
             $values = array_values($values);
         }
 
-        if (!is_string($header) || empty($header)) {
-            throw new \InvalidArgumentException('Header name must be strings');
+        if (!is_string($header) || preg_match("@^[!#$%&'*+.^_`|~0-9A-Za-z-]+$@", $header) !== 1) {
+            throw new \InvalidArgumentException('Header name must be an RFC 7230 compatible string.');
         }
 
         foreach ($values as &$v) {
             if (is_numeric($v)) {
                 $v = (string) $v;
-            } elseif (!is_string($v)) {
-                throw new \InvalidArgumentException('Header values must be strings');
+            } elseif (!is_string($v) || preg_match("@^[ \t\x21-\x7E\x80-\xFF]*$@", $v) !== 1) {
+                throw new \InvalidArgumentException('Header values must be RFC 7230 compatible strings.');
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/MessageTrait.php
```diff
@@ -105,35 +89,12 @@ public function withHeader($header, $value): self
...
     {
-        if (!is_array($value)) {
-            $value = [$value];
-        } elseif (!empty($value)) {
-            $value = array_values($value);
-        } else {
-            throw new \InvalidArgumentException('Header values must be strings');
-        }
-
         if (!is_string($header) || empty($header)) {
```
Zegnat: Why do we still have validation here when this should be handled by `validateAndTrimHeader`?

src/MessageTrait.php
```diff
@@ -196,21 +152,39 @@ private function setHeaders(array $headers): void
...
+        if (!is_array($values)) {
+            $values = [$values];
+        } elseif (empty($values)) {
+            throw new \InvalidArgumentException('Header values must be strings, empty array given.');
+        } else {
+            // Non empty array
+            $values = array_values($values);
+        }
+
+        if (!is_string($header) || empty($header)) {
```
Zegnat: Should we also check that the header name is allowed per RFC 7230? More specifically this means it consists of 1 or more `tchar` exclusively:

```
tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
               / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
               / DIGIT / ALPHA
DIGIT          = %x30-39
ALPHA          = %x41-5A / %x61-7A
```

This will also make sure we never accidentally add a header name ending in whitespace, which is [explicitly forbidden](https://tools.ietf.org/html/rfc7230#section-3.2.4):

> No whitespace is allowed between the header field-name and colon. In the past, differences in the handling of such whitespace have led to security vulnerabilities in request routing and response handling.

src/MessageTrait.php
```diff
@@ -196,21 +152,39 @@ private function setHeaders(array $headers): void
...
+            throw new \InvalidArgumentException('Header name must be strings');
+        }
+
+        foreach ($values as &$v) {
+            if (is_numeric($v)) {
+                $v = (string) $v;
+            } elseif (!is_string($v)) {
```
Zegnat: Should we also check that the header value is allowed per RFC 7230? More specifically this means it consists of 0 or more `field-content` exclusively (IMHO we can ignore `obs-fold` as it was [deprecated in most use-cases](https://tools.ietf.org/html/rfc7230#section-3.2.4)):

```
field-value    = *( field-content / obs-fold )
field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
field-vchar    = VCHAR / obs-text
obs-text       = %x80-FF
VCHAR          = %x21-7E
```

Zegnat:
I am happy to commit to the issue-40 branch myself. But don’t want to step on your toes 😄

src/MessageTrait.php
Nyholm: That would be great! Can you make that change?

src/MessageTrait.php
Nyholm: True. It should be removed.

Nyholm: No, If $header is not a string (ie an object) it will fail when we try to use it as an array key.

Nyholm:
Please do commit to this branch. I do not mind at all.

src/MessageTrait.php
```diff
@@ -105,35 +89,8 @@ public function withHeader($header, $value): self
...
     {
-        if (!is_array($value)) {
-            $value = [$value];
-        } elseif (!empty($value)) {
-            $value = array_values($value);
-        } else {
-            throw new \InvalidArgumentException('Header values must be strings');
-        }
-
-        if (!is_string($header) || empty($header)) {
```
Nyholm: Cannot remove this. 
What if $header is an object? We cannot use objects as array keys.

src/MessageTrait.php
```diff
@@ -163,21 +170,21 @@ private function validateAndTrimHeader($header, $values): array
         if (!is_array($values)) {
             $values = [$values];
         } elseif (empty($values)) {
-            throw new \InvalidArgumentException('Header values must be strings, empty array given.');
+            throw new \InvalidArgumentException('Header values must be an array of strings, empty array given.');
```
Nyholm: Lets do: 
> Header values must be a string or an array of strings, empty array given.

src/MessageTrait.php
Zegnat: Yeah, I missed your comment and for some reason I didn’t realise that was what was happening. I’ll put it back!

src/MessageTrait.php
Zegnat: Done!

src/MessageTrait.php
```diff
@@ -89,6 +89,10 @@ public function withHeader($header, $value): self
 
     public function withAddedHeader($header, $value): self
     {
+        if (!is_string($header) || preg_match("@^[!#$%&'*+.^_`|~0-9A-Za-z-]+$@", $header) !== 1) {
```
Nyholm: Do we **really** care about RFC 7230 here? That check is done twice now. It would be faster to just check if it is empty or not.

Zegnat: `empty` is definitely incorrect, as it turns down `"0"`. (No strict typing on that function.) We could do a `strlen`.

Nyholm: That is right. 

Could we just do `!is_string($header) || '' === $header`?

Zegnat: He, I just pushed with `strlen`. But you are right that a strict comparison to `''` would work too. I have no data to suggest either of those two is faster. Pick whichever, I guess?

Nyholm:
Without knowing.. I would assume that `'' === $header` is constant time. Doing a strlen would have to count all characters which *could* be linear time.

Nyholm:
We could ignore travis. I think it will be all green when all open PRs are merged to master.

Zegnat:
I don’t know how PHP internally handles strings or how `strlen` counts their size. But my own [super rigorous and in-depth testing](https://gist.github.com/Zegnat/b57dddd2261260fc0b408ad42d2d4af3) has the strict comparison win in speed on my own computer. So I’ll rewrite to that.

I don’t know why Travis is failing though, it doesn’t seem to have anything to do with the message trait …

Nyholm:
I got the same result on my computer. Thank yo

Nyholm:
Thank you. Could you review the other PRs. I'm happy to merge them all now but I want your input first. 

Lets merge and fix potential bugs in master =)

PR #49, opened 2018-07-23
Added changelog for 0.4.0

Nyholm:
Release is blocked by #48 and #55.


Zegnat:
Per #58 we may want to put the unreleased version number to 1.0.0, which is a nice commitment to give. Otherwise, this looks good and seems to cover the major updates. :+1:

Nyholm:
I would like to be a bit more "careful". To have 0.4.0 unchanged for a month or two and then tag 1.0.0 if nothing comes up

PR #54, opened 2018-07-28
Document simple usage

### LINKED ISSUE #53: Documentation
I really like this library as it's small and fast. While it's not popular as Zend Diactoros or Guzzle, it strictly follows the PSR standards, which I really appreciate.

However, one thing that makes me wondering from use it or not is the documentation. There's *nothing*, even a "hello world" example. Other libraries do have. So, I'd suggest adding some basic usage in the README to let developers how to use it.

Thanks

Nyholm:
This will fix #53


rilwis:
Great! Thanks a lot!

PR #55, opened 2018-07-28
Moved non-psr7 functions to a separate package

### LINKED ISSUE #44: Separate class specifically for building ServerRequest instance of the current HTTP request.
As I documented in #43, there isn’t an automatic way (defined by PSR-17) to take the current HTTP request as seen by PHP (combination of superglobals, `php://input`, etc.) and make it a `ServerRequest` instance. This is by design.

I’d propose building a separate helper class following [an idea initially mentioned on the PSR-17 repo](https://github.com/http-interop/http-factory/issues/31#issuecomment-300560775):

``` php
use ServerRequestCreator\Creator;

$creator = new Creator(new PsrServerRequestFactory());

$serverRequest = $creator->fromGlobals([
    'useMethodOverride' => true,
    'useIISRewriteModule' => true
]);
```

I am not so sure about the options array used there, or even the names, but it would make sense to me to offer this as part of the package. Simply because people expect it. Maybe something akin to:

``` php
use Nyholm\Psr7\Factory\ServerRequestFactory;
use Nyholm\Psr7\Factory\StreamFactory;
use Nyholm\Psr7\Helper\ServerRequestCreator;

$creator = new ServerRequestCreator(
    new ServerRequestFactory(),
    new StreamFactory()
);

$serverRequest = $creator->fromGlobals();
```

Nyholm:
This will fix #44. See https://github.com/Nyholm/psr7-server

This builds on top of #54 (my bad). I will rebase when #54 is merged.


Nyholm:
(doesn't it feel great to remove code 😄 )

Nyholm:
I originally put server request separate because of all this code. But since we are removing this code now, I think the next pr should merge the serverrequest to psr17factory

Nyholm:
Change my mind. You are very correct. It is easier to do it now. 

Feel free to merge if you like it.

Nyholm:
(I just rebased this PR.)

Nyholm:
@zegnat, could you review please?

Nyholm:
Since HTTPlug supports PHP5 we cannot use the same class for HTTPlug's UriFactory and PSR17's UriFactory.

Zegnat:
> 

Aah, I was looking at the `use` statements up top and missed the namespace on the `implements`! In that case: LGTM :+1:

Nyholm:
Thank you for the review

PR #60, opened 2018-08-01
Move StreamFactory::copyToStream to (private) UploadedFile

Nyholm:


Nyholm:
This is the last public method Ive found that is not described in an interface.

Zegnat:
This looks right. The only others I have found were `Stream::__destruct`, `Stream::createFromResource`, and `Stream::create`.

I don’t see how `Stream::__destruct` could ever be a problem. But the other two could theoretically get redefined by PSR in the future. (Though that sounds extremely unlikely.)

Nyholm:
Good. Excellent. Lets do this!

PR #62, opened 2018-08-01
Merge HTTPlug (PHP-HTTP) factories into single class

Zegnat:
Basically more streamlining. This addresses my own confusion with the existence of separate factory files, as mentioned in https://github.com/Nyholm/psr7/pull/55#issuecomment-409355463.

Based on #60 and should be merged after.

With this change I also noticed we are calling our own factories within methods of the PSR-7 classes… why?


Nyholm:
Hm. Moving them to one factory makes sense. I agree. However, that will mess up php-http/discovery. They rely on these classes. We *could* update the discovery strategy. That would work if we assume everyone updates their dependencies. 

What do you think? Is it worth changing these classes?
I think I would rather leave them and deprecate them as soon as 1.0 is released. I know that HTTPlug will deprecate these factories "soon" as well.

Zegnat:
Rebased this now that #60 was merged.

> What do you think? Is it worth changing these classes?

On the one hand I would love for the 1.0.0 release to have a minimal amount of files and think it is weird that we can’t do that because of a completely unrelated package (minus your involvement in HTTPlug in general).

On the other hand, if HTTPlug is going to pull the rug from [php-http/message-factory](https://github.com/php-http/message-factory) anyway, maybe we should cut our losses and look forward to a 2.0.0 release completely without these classes instead?

src/Factory/PhpHttpFactory.php
```diff
@@ -0,0 +1,51 @@
...
+use Nyholm\Psr7\Stream;
+use Nyholm\Psr7\Uri;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ * @author Martijn van der Ven <martijn@vanderven.se>
+ */
+final class PhpHttpFactory implements \Http\Message\MessageFactory, \Http\Message\StreamFactory, \Http\Message\UriFactory
```
Nyholm: Lets name it to HttplugFactory

Zegnat: Done.

Zegnat:
Wow did my rebase get confused after you squashed #64 …

PR #63, opened 2018-08-01
Removed unneeded call to factory

Nyholm:
This is a simple fix


Zegnat:
The trait is no longer using the factory at all after this, so it is fine to drop the `use` statement. (That saves another loc :wink:)

Nyholm:
Thank you. I forgot that.

PR #64, opened 2018-08-01
Duplicated logic to avoid creating new instance of StreamFactory

Nyholm:
Im not sure what to do here. Is this a good solution? 

The logic is copied from `StreamFactory`. 
The alternative would be to create a new public method on `Stream` and **move** the code from `StreamFactory`.

### DIFF
--- a/src/Factory/Psr17Factory.php
+++ b/src/Factory/Psr17Factory.php
@@ -46,12 +46,12 @@ final class Psr17Factory implements RequestFactoryInterface, ResponseFactoryInte
 
     public function createStreamFromFile(string $filename, string $mode = 'r'): StreamInterface
     {
-        return Stream::createFromResource(fopen($filename, $mode));
+        return Stream::create(fopen($filename, $mode));
     }
 
     public function createStreamFromResource($resource): StreamInterface
     {
-        return Stream::createFromResource($resource);
+        return Stream::create($resource);
     }
 
     public function createUploadedFile(StreamInterface $stream, int $size = null, int $error = \UPLOAD_ERR_OK, string $clientFilename = null, string $clientMediaType = null): UploadedFileInterface
--- a/src/Factory/StreamFactory.php
+++ b/src/Factory/StreamFactory.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Nyholm\Psr7\Factory;
 
 use Nyholm\Psr7\Stream;
-use Psr\Http\Message\StreamInterface;
 
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
@@ -14,14 +13,6 @@ final class StreamFactory implements \Http\Message\StreamFactory
 {
     public function createStream($body = null)
     {
-        if ($body instanceof StreamInterface) {
-            return $body;
-        }
-
-        if ('resource' === gettype($body)) {
-            return Stream::createFromResource($body);
-        }
-
         return Stream::create(null === $body ? '' : $body);
     }
 }
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Nyholm\Psr7;
 
-use Nyholm\Psr7\Factory\StreamFactory;
 use Psr\Http\Message\StreamInterface;
 
 /**
@@ -115,7 +114,7 @@ trait MessageTrait
     public function getBody(): StreamInterface
     {
         if (!$this->stream) {
-            $this->stream = (new StreamFactory())->createStream('');
+            $this->stream = Stream::create('');
         }
 
         return $this->stream;
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -193,4 +193,17 @@ trait MessageTrait
             return trim($value, " \t");
         }, $values);
     }
+
+    private function createStream($body = null): StreamInterface
+    {
+        if ($body instanceof StreamInterface) {
+            return $body;
+        }
+
+        if ('resource' === gettype($body)) {
+            return Stream::createFromResource($body);
+        }
+
+        return Stream::create(null === $body ? '' : $body);
+    }
 }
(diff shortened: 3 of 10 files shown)


Zegnat:
I am not sure what the best solution is either. I think it is right to not use `StreamFactory`, especially because that isn’t the PSR-17 implementation to begin with (and may be removed or deprecated).

Couldn’t `$body` just be type-hinted as a `StreamInterface` in the `__construct`? Or is that too limiting?

Nyholm:
I think that is too limiting. I like the signature of the StreamFactory. It is the same signature Guzzle is using for `stream_for` or whatever their method is called. 

We can add a public method to the Stream later if we want to.

Zegnat:
After another thought, I really like the `createStream` method. We could use that to validate the `$body` value and throw `InvalidArgumentException` if something is submitted that does not match our signature (`string|null|resource|StreamInterface`). A little bit like how [validation for header values](https://github.com/Nyholm/psr7/pull/48) has been abstracted:

```php
        if (is_string($body)) {
            return Stream::create($body);
        }

        throw new \InvalidArgumentException('Body must be a string, resource, or StreamInterface.');
```

(No need for the `null` check within the `Stream::create` call, as we check that before calling `MessageTrait::createStream` anyway.)

Thoughts?

Nyholm:
Sure, So we basically put the logic from `StreamFactory::create` to `Stream::create`?

Zegnat:
> So we basically put the logic from `StreamFactory::create` to `Stream::create`?

Yeah. Feels a little weird still. But because they basically use the same signature that’s what you end up with. Unless you peg one to the other, but that is weird as well. (Internally depending on a factory like we do now?)

> We can add a public method to the Stream later if we want to.

This would be an alternative too, something like `Stream::createFromUntyped`? But I think that is slightly worse as you end up having to support another non-strictly typed public method.

Adding a private helper method to the trait does not involve future work.

Zegnat:
How about truly mirroring `MessageTrait::setHeaders` with a `MessageTrait::setStream`?

```php
    private function setStream($body): void
    {
        if ('' !== $body && null !== $body) {
            if ($body instanceof StreamInterface) {
                $this->stream = $body;
            } elseif ('resource' === gettype($body)) {
                $this->stream = Stream::createFromResource($body);
            } elseif (is_string($body)) {
                $this->stream = Stream::create($body);
            } else {
                throw new \InvalidArgumentException('Body must be a string, resource, or StreamInterface.');
            }
        }        
    }
```

Then all `Request::__construct`, `Response::__construct`, and `ServerRequest::__construct`, need to do is call `$this->setStream($body)` and forget about the rest.

Nyholm:
> How about truly mirroring MessageTrait::setHeaders with a MessageTrait::setStream?

That is exactly what tis PR is doing... right?

>  So we basically put the logic from StreamFactory::create to Stream::create

Maybe I like that more. To update `Stream::create` instead.

Zegnat:
> That is exactly what tis PR is doing... right?

Yes. Except I’m saying we move the outer `if` into it and added proper exception throwing.

> Maybe I like that more. To update `Stream::create` instead.

So that would make `Stream::create`’s argument no longer strictly typed, right? In that case, it might be better to move it to `Stream::_construct`?

```php
/**
 * @param StreamInterface|resource|string $body Body of the stream.
 **/
public function __construct($body)
```

Nyholm:
> So that would make Stream::create’s argument no longer strictly typed, right?

Yes, lets do that. 

> In that case, it might be better to move it to Stream::_construct?

Na, not really. Bacause the constructor cannot handle if someone passes an `StreamInterface`.

Nyholm:
I also made `Stream::createFromResource` private. That made the API simpler.

Zegnat:
Are we using `Stream::createFromResource` anywhere outside of `Stream::create`? I didn’t spot any other places. If we aren’t, why not get rid of it completely? Just need to do a small reshuffle of the `if`s so we convert string `$body` values to a resource first:

```diff
diff --git a/src/Stream.php b/src/Stream.php
index 459dad9..e197dde 100644
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -50,26 +50,6 @@ final class Stream implements StreamInterface
     {
     }
 
-    /**
-     * @param resource $resource
-     */
-    private static function createFromResource($resource): self
-    {
-        if (!is_resource($resource)) {
-            throw new \InvalidArgumentException('Stream must be a resource');
-        }
-
-        $obj = new self();
-        $obj->stream = $resource;
-        $meta = stream_get_meta_data($obj->stream);
-        $obj->seekable = $meta['seekable'];
-        $obj->readable = isset(self::$readWriteHash['read'][$meta['mode']]);
-        $obj->writable = isset(self::$readWriteHash['write'][$meta['mode']]);
-        $obj->uri = $obj->getMetadata('uri');
-
-        return $obj;
-    }
-
     /**
      * Creates a new PSR-7 stream.
      *
@@ -81,20 +61,26 @@ final class Stream implements StreamInterface
      */
     public static function create($body = ''): StreamInterface
     {
-        if ('resource' === gettype($body)) {
-            return self::createFromResource($body);
-        }
-
         if ($body instanceof StreamInterface) {
             return $body;
         }
 
         if (is_string($body)) {
             $resource = fopen('php://temp', 'rw+');
-            $stream = self::createFromResource($resource);
-            $stream->write($body);
+            fwrite($resource, $body);
+            $body = $resource;
+        }
 
-            return $stream;
+        if ('resource' === gettype($body)) {
+            $obj = new self();
+            $obj->stream = $body;
+            $meta = stream_get_meta_data($obj->stream);
+            $obj->seekable = $meta['seekable'];
+            $obj->readable = isset(self::$readWriteHash['read'][$meta['mode']]);
+            $obj->writable = isset(self::$readWriteHash['write'][$meta['mode']]);
+            $obj->uri = $obj->getMetadata('uri');
+
+            return $obj;
         }
 
         throw new \InvalidArgumentException('First argument to Stream::create() must be a string, resource or StreamInterface.');
```

Nyholm:
Yeah. That is better. 

Great :+1:

Zegnat:
All checks passed, looks good to me :+1:

Nyholm:
That last commit was not needed. That would have been fixed by a rebase. 

Thank you for completing this PR.

PR #66, opened 2018-08-02
Mark our traits as internal.

Nyholm:


src/MessageTrait.php
```diff
@@ -11,6 +11,8 @@
  *
  * @author Michael Dowling and contributors to guzzlehttp/psr7
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ *
+ * @internal should not be used outside of Nyholm/Psr7 as it does not fall under out BC promise.
```
Zegnat: Typo: “out” should be “our”. Same in `RequestTrait`.

Nyholm: Thank you

Nyholm:
Thank you for the review

PR #67, opened 2018-08-02
Shorten multiple use statements

Nyholm:


Zegnat:
I have zero preference here. I guess this saves some LOC? Otherwise [grouping](https://secure.php.net/manual/en/language.namespaces.importing.php#language.namespaces.importing.group) could be used to save a few bytes.

Nyholm:
Indeed =)

I did not know about grouping. I've updated the PR

Nyholm:
Awesome. 1030 lines now. =)

Nyholm:
Sure, that make sense to be constant. I'll add that line now

Zegnat:
Went through all the files in src again, and looks like all classes are either in grouped `use` statements or prefixed with `\` for root now. Nice :+1:

Nyholm:
Great. Thanks!

PR #70, opened 2018-08-02
Add MIT license

Zegnat:
This is probably a good idea before the 1.0.0 release?


Nyholm:
Oh, I was surprised there weren't none. Thanks!

PR #72, opened 2018-08-02
cs fixer does not support our use statements.

Nyholm:
Even with `single_import_per_statement: false`


Zegnat:
Can’t comment on this as CS isn’t part of the composer.json package and I haven’t been testing with it. Surprising though.

Issue #71, opened 2018-08-02
Scope all constants to root, or not?


Zegnat:
When I was checking `\` prefixes for #67 I saw the factory uses `\UPLOAD_ERR_OK`:

https://github.com/Nyholm/psr7/blob/36870acbf2ef440db6a072509afc92f5d4155d77/src/Factory/Psr17Factory.php#L41

Every other time constants are used, we seem to leave it up to PHP to figure out we want them from outside our own namespace. Here is `UPLOAD_ERR_OK`:

https://github.com/Nyholm/psr7/blob/90807bc10796287265f726e942bcbb2272fbc5b6/src/UploadedFile.php#L137

Is there some way to make StyleCI check this? Is it even important?

Nyholm:
Yes. Lets do `\` before each constant. 

FYI: https://github.com/php-fig/http-factory/issues/3

PR #74, opened 2018-08-02
Pin specific functions to root namespace for optimisation

Zegnat:
Further change coming out of #71, this specifies certain functions as coming from the root namespace.

Only functions on [this list of optimised functions](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/b38da665fb73a25cdbc8311b439b7298fb50911c/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L347-L394) have been prefixed.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -90,7 +90,7 @@ trait MessageTrait
 
     public function withAddedHeader($header, $value): self
     {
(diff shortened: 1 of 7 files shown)


Nyholm:
Excellent

PR #75, opened 2018-08-02
Code style fixes

Nyholm:
These fixes uses the new php_cs. 

They are just for decoration. No performance boost.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,7 +1,15 @@
 <?php
 
 $config = PhpCsFixer\Config::create()
-
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => array('syntax' => 'short'),
+        'native_function_invocation' => true,
+        'ordered_imports' => true,
+        'declare_strict_types' => true,
+        'single_import_per_statement' => false,
+    ])
     ->setFinder(
         PhpCsFixer\Finder::create()
             ->in(__DIR__.'/src')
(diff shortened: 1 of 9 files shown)


Zegnat:
Can we get PHP_CodeSniffer added to the composer.json? Maybe also add it as a `script`? Cf. [PHP League Skeleton](https://github.com/thephpleague/skeleton/blob/04ab224bf800e6236e1ead3910b2a2aaa02a15a5/composer.json#L38-L39):

```json
        "check-style": "phpcs -p --standard=PSR2 --runtime-set ignore_errors_on_exit 1 --runtime-set ignore_warnings_on_exit 1 src tests",
        "fix-style": "phpcbf -p --standard=PSR2 --runtime-set ignore_errors_on_exit 1 --runtime-set ignore_warnings_on_exit 1 src tests"
```

Nyholm:
I do not mind having it as a composer script. But Im not super exited to put PHP_CodeSniffer as a dependency. It is just a tool, not something we need to run our code or our tests.

Zegnat:
I am one of those strange people who doesn’t have PHPUnit or PHP_CodeSniffer installed globally on his system, so I might just be projecting whatever I feel like is the path of least resistance. Haha.

If you don’t feel like it belongs together with tests (even though all PRs get a style check by StyleCI already), don’t add it to the file. I don’t feel strongly either way, so if you aren’t excited lets ignore it :+1:

PR #78, opened 2018-08-12
Reduce the execution path to speed up performance

Nyholm:
screen shot 2018-08-11 at 14 55 36

See tests here https://github.com/Nyholm/http-client-benchmark


Zegnat:
Several functions in here have not been “rooted” (e.g. `is_numeric` rather than `\is_numeric`) while #75 shows they should be … did we ever get some consensus on this?

What tool did you use in that screenshot? I was recently looking for some benchmarking tools for [a different project](https://github.com/indieweb/php-mf2).

Code looks good and seems like it does exactly what the old code did. So :+1: on merge from me.

Nyholm:
They should be rooted. 

I use blackfire. That is a great tool =)

I'll merge this and make sure to rebase #75.

PR #84, opened 2018-08-31
Handle fopen failing in createStreamFromFile: throw a RuntimeException.

Zegnat:
I noticed this because of Nyholm/psr7-server#20. It looked like we were assuming `fopen` to always succeed rather than checking, this would even leek the PHP E_WARNING to the end-user.

This PR will suppress the E_WARNING (even the PHP manual tells us we might want to do that) and explicitly check for failure. If the file couldn’t be opened, it throws a `RuntimeException` in accordance with the PSR-17 definition.

### DIFF
--- a/src/Factory/Psr17Factory.php
+++ b/src/Factory/Psr17Factory.php
@@ -30,7 +30,12 @@ final class Psr17Factory implements RequestFactoryInterface, ResponseFactoryInte
 
     public function createStreamFromFile(string $filename, string $mode = 'r'): StreamInterface
     {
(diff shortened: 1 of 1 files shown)


Zegnat:
This PR does not include any tests, because we do not include any factory tests with this library. Instead we rely on [shared http-interop tests](https://github.com/http-interop/http-factory-tests). I have opened an issue there to track further work on testing StreamFactory exceptions: http-interop/http-factory-tests#35.

Zegnat:
PR for tests has been created: http-interop/http-factory-tests#36.

PR #90, opened 2018-09-15
Pattern readability

rancoud:
For better readability, I extract the regex to the top of the class in two static properties like in `Uri`


Nyholm:
Thank you for this PR. 

Im not sure about this. I get the argument for DRY. But Im not sure it improves readability. It moves the regex so far away from where it is being used.

rancoud:
Ok, no problem :)

PR #91, opened 2018-09-15
Protocol validation

rancoud:
Following the RFC, HTTP version will be validate against a method.

It will throw an exception if invalid.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -36,6 +36,8 @@ trait MessageTrait
 
     public function withProtocolVersion($version): self
     {
+        $this->validateProtocolVersion($version);
+
         if ($this->protocol === $version) {
             return $this;
         }
@@ -198,5 +200,14 @@ trait MessageTrait
         }
 
         return $returnValues;
+    }
+
+    private function validateProtocolVersion(string $protocolVersion): string
+    {
+        if (!\in_array($protocolVersion, ['0.9', '1.0', '1.1', '2'], true)) {
+            throw new \InvalidArgumentException('Protocol Version must be 0.9 or 1.0 or 1.1 or 2');
+        }
+
+        return $protocolVersion;
     }
 }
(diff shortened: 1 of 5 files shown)


Nyholm:
Thank you for this PR. Are you sure about this? Could you give me a reference to the mentioning of this in the PSR?

rancoud:
I have no reference about this in PSR.

But in the RFC 7230 they talked about [3 versions (0.9, 1.0, 1.1)](https://tools.ietf.org/html/rfc7230#appendix-A) and there is another [RFC 7540 about 2.0](https://tools.ietf.org/html/rfc7540)

So I suppose, it will be interesting to limit this value.

Zegnat:
[PSR-7](https://www.php-fig.org/psr/psr-7/) tries to represent messages per [RFC 7230](https://tools.ietf.org/html/rfc7230). And while the linked appendix talks about known versions, I think the message trait should be able to support everything a HTTP message per the RFC supports.

Section [2.6. Protocol Versioning](https://tools.ietf.org/html/rfc7230#section-2.6) has the full write-up, describing when major and minor versions are to be bumped. This section doesn’t limit the version to anything specific. Other than it always being a single digit on both sides of the period.

The section also describes how senders and receivers should handle versions that are higher than whatever they support. In fact, this means there shouldn’t be any harm done if someone wants to start using `1.2` for whatever reason. Blocking that as an invalid value in this library would mean you couldn’t construct a `1.2` request to be sent with a PSR-18 client. Why should the HTTP message implementation be a limiter there?

For less reading, check [Appendix B](https://tools.ietf.org/html/rfc7230#appendix-B) for the definition of `HTTP-version`: `HTTP-name "/" DIGIT "." DIGIT`.

I’d be for limiting the protocol to `\d\.\d`, but limiting to specific values seems unnecessary.

rancoud:
Ok, no problem.

PR #92, opened 2018-09-15
Improve coverage

rancoud:
Add tests for improving coverage.

### DIFF
--- a/tests/RequestTest.php
+++ b/tests/RequestTest.php
@@ -201,4 +201,12 @@ class RequestTest extends TestCase
         $r = $r->withHeader('Foo', '');
         $this->assertEquals([''], $r->getHeader('Foo'));
     }
+
+    public function testCannotHaveEmptyHeaderValues()
+    {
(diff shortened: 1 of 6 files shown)


tests/RequestTest.php
```diff
@@ -201,4 +201,12 @@ public function testCanHaveHeaderWithEmptyValue()
         $r = $r->withHeader('Foo', '');
         $this->assertEquals([''], $r->getHeader('Foo'));
     }
+
+    public function testCannotHaveEmptyHeaderValues()
```
Zegnat: Is there a better name for this? The real error being tested here is that `null` isn’t a valid string. Empty header values (that is: empty strings) are [valid header values](https://github.com/php-http/psr7-integration-tests/pull/30).

rancoud: I rename it to `testCannotHaveHeaderValueWithNullValue`

Zegnat: :+1: Thanks! Mostly because I know I was going to get confused about headers again in a month time otherwise 😉

PR #97, opened 2018-11-23
Made codes more readable, prettier

AliN11:
I tried to make the codes become more standard, readable and prettier. Although it could be better yet.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Nyholm\Psr7;
 
+use InvalidArgumentException;
 use Psr\Http\Message\UriInterface;
 
 /**
@@ -48,7 +49,7 @@ final class Uri implements UriInterface
     {
         if ('' !== $uri) {
             if (false === $parts = \parse_url($uri)) {
-                throw new \InvalidArgumentException("Unable to parse URI: $uri");
+                throw new InvalidArgumentException("Unable to parse URI: $uri");
             }
 
             $this->applyParts($parts);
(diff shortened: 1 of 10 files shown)


AliN11:
Do you know why Travis check fails?

AliN11:
Do you know why Travis check fails?

src/Uri.php
```diff
@@ -269,13 +280,16 @@ private static function createUriString(string $scheme, string $authority, strin
      */
     private static function isNonStandardPort(string $scheme, int $port): bool
     {
-        return !isset(self::$schemes[$scheme]) || $port !== self::$schemes[$scheme];
+        $scheme_not_set = !isset(self::$schemes[$scheme]);
+        $is_invalid_port = $port !== self::$schemes[$scheme];
```
Zegnat: This is the line that breaks the code. Our tests correctly catch this, thus Travis fails.

Your two lines of code are not the same as the old one line of code. For one line of code, the second part of the “or” statement does not run when the first part evaluates as false. We specifically make use of this by checking if our array contains an item with the key `$scheme` before we access it.

Your changed code is always going to try and access the array item with the key `$scheme`. When you ask PHP to do that and there is no such item in the array, it will fail.

I think – though this is without testing or guarantees – the actual equivalent to the old code would be:

```php
if (false === isset(self::$schemes[$scheme])) {
    return true;
} else if ($port !== self::$schemes[$scheme]) {
    return true;
}
return false;
```

This means the second check will only be done when the first one didn’t pass.

Zegnat:
This seems fine even with our current style guide. It also doesn’t look like there are a lot of functional changes in here that would affect performance. I do know that @Nyholm likes LOC as a measurement for this library so I am not sure extra linebreaks are going to get merged in.

src/Uri.php
AliN11: Thanks man. I rolled back that method to the initial state. So Travis check passed successfully.

AliN11:
@Zegnat Thanks for reviewing. However, your nice package could be nicer :wink:

Nyholm:
Hey. 

Thank you for the PR and for the review. 
As @Zegnat says, This does not bring any functional changes and "pretty" is highly subjective. 

I appreciate this contribution but I see no reason to merge at this point.

PR #98, opened 2018-12-27
Stream: seek to the begining of the string on initialization

lyrixx:
This is useful in test env where we usually 'mock' the response

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.tmp
 /build/
 /composer.lock
 /phpunit.xml


Zegnat:
This would be a breaking change as it affects any subsequent calls to `write()`, `read()`, etc. So I’d like to hold of on merging this pending further discussion. Also because `Stream::create` for this package is directly mapped to the PSR-17 factory implementation.

This has spurred me to do some research showing that [PSR-17 implementations disagree on where the file position indicator ends up](https://github.com/Zegnat/php-psr17-test-cursor). I have started [a discussion on the PHP-FIG mailing list](https://groups.google.com/forum/?fromgroups#!topic/php-fig/S5YIw-Pu1yM). Hopefully we can get some eyes on this and have an expected outcome formalised in the [official unit tests](https://github.com/php-fig/http-factory).

For now I think the best advice possible for compatibility is to always call `rewind()` when you are working with a newly created stream. Or at least check where you are with `tell()`.

lyrixx:
For now I'm using `__toString` method in my code to "fix" this issue
https://github.com/Nyholm/psr7/blob/867ce43052cb60e32001a8a0aa825f1830882ad8/src/Stream.php#L98-L109

Zegnat:
> For now I'm using `__toString` method in my code to "fix" this issue

That may not only be a fix, but just the most correct way of using PSR-7 objects.

[PSR-7](https://www.php-fig.org/psr/psr-7/) defines `StreamInterface::__toString()` to return “all data from the stream into a string, from the beginning to end”, while `StreamInterface::getContents()` returns “the remaining contents” (not necessarily from the start). None of that is particularly obvious though, and in certain environments people may even dislike depending on magic methods such as `__toString()`.

Nevertheless I continued researching where different implementations put the cursor (and thus the start of `getContents()`) and sadly concluded that there is [a near 50/50 split](https://groups.google.com/d/msg/php-fig/S5YIw-Pu1yM/RKbk6CTdCgAJ). Without a clear right answer, I do not think this PR can be merged. I would not like to introduce a breaking change and later have to revert it again if the PHP-FIG Working Group decides that putting the cursor at the end of the string is correct.

This also means that if you want to reserve the ability to switch PSR-7/PSR-17 implementations in the future, it is best not to depend on a fixed cursor position from `StreamFactoryInterface::createStream()`.

If you feel strongly for or against a certain cursor position, I would love to read your opinions on the PHP-FIG mailing list as well. Maybe a case is to be made around mocking. I have found the PHP-FIG to be plenty open to outsider perspectives! (Like mine, for that matter.)

PR #105, opened 2019-02-14
CS fixes

Nyholm:


src/Request.php
```diff
@@ -16,11 +16,11 @@
     use RequestTrait;
 
     /**
-     * @param string $method HTTP method
-     * @param string|UriInterface $uri URI
-     * @param array $headers Request headers
-     * @param string|null|resource|StreamInterface $body Request body
-     * @param string $version Protocol version
+     * @param string                               $method  HTTP method
```
samdark: This won't be parse-able via IDEs. At least PhpStorm.

src/Uri.php
```diff
@@ -309,7 +309,7 @@ private function filterQueryAndFragment($str): string
             throw new \InvalidArgumentException('Query and fragment must be a string');
         }
 
-        return \preg_replace_callback('/(?:[^' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . '%:@\/\?]++|%(?![A-Fa-f0-9]{2}))/', [__CLASS__, 'rawurlencodeMatchZero'], $str);
+        return \preg_replace_callback('/(?:[^'.self::CHAR_UNRESERVED.self::CHAR_SUB_DELIMS.'%:@\/\?]++|%(?![A-Fa-f0-9]{2}))/', [__CLASS__, 'rawurlencodeMatchZero'], $str);
```
samdark: And this is against formatting PSR.

src/Request.php
Nyholm: Yeah, I saw that. I will figure out what php-cs-fixer rule that disables this.

src/Uri.php
Nyholm: These changes are automatic. I need to fine the correct php-cs-fixer rule to change this. 

I do not want to undo your changes.

Nyholm:
For future reference: If you run `php-cs-fixer fix -vvv` it will say **why** they are fixing some files. It makes it easier to find the correct rule in the docs.

Nyholm:
I have addressed the issues. 

Travis is red because it claims it is a BC break. 
I did not change the behavior of `MessageTrait::$stream`, it was the documentation that was wrong the in the first place.

PR #107, opened 2019-02-16
Added more tests and scrutnizer fixes

Nyholm:


src/UploadedFile.php
```diff
@@ -40,7 +40,7 @@
     /** @var bool */
     private $moved = false;
 
-    /** @var int|null */
+    /** @var int */
```
Nyholm: It is actually impossible to create an UploadedFile with `$size = null`

src/UploadedFile.php
```diff
@@ -212,34 +212,17 @@ public function getClientMediaType(): ?string
      *
      * @param StreamInterface $source Stream to read from
      * @param StreamInterface $dest Stream to write to
-     * @param int $maxLen Maximum number of bytes to read. Pass -1
-     *                    to read the entire stream
      *
      * @throws \RuntimeException on error
      */
-    private function copyToStream(StreamInterface $source, StreamInterface $dest, $maxLen = -1): void
+    private function copyToStream(StreamInterface $source, StreamInterface $dest): void
```
Nyholm: We are never using the third parameter, That is why I could remove the dead code

tests/StreamTest.php
```diff
@@ -19,7 +19,7 @@ public function testConstructorInitializesProperties()
         $this->assertTrue($stream->isWritable());
         $this->assertTrue($stream->isSeekable());
         $this->assertEquals('php://temp', $stream->getMetadata('uri'));
-        $this->assertInternalType('array', $stream->getMetadata());
+        $this->assertIsArray($stream->getMetadata());
```
Nyholm: `assertInternalType` is deprecated

Nyholm:
Im done with this PR now.

Zegnat:
I don’t understand the Travis error on this one… It seems to complain about `MessageTrait` but I don’t see you changing `MessageTrait`?

I haven’t used PHPStan before so this will be interesting to check out. Overall all the code changes look good to me!

Nyholm:
PHPStan is pretty cool. We do get a few false negatives but that is fine. 

I added the docs on `MessageTrait::$stream` to allow null. We already did allow it but it was not documented. That is the "BC break". 

It is not an actual BC break for two reasons: 
- It is a private property on our "internal trait"
- The behavior of the code was never changed. 

Thank you for the review

PR #112, opened 2019-04-01
Typo

Spomky:


Zegnat:
@Nyholm do you know why this isn’t getting passed Travis (and thus unmergable)?

Nyholm:
Hm. No. I’ll have a look as soon as I land (if battery allows it)

PR #115, opened 2019-04-15
Change minimal port number to 0 (unix socket)

Tracerneo:
Listening on an unix socket reports port number 0, which is interpreted as an invalid argument.

### DIFF
--- a/tests/UriTest.php
+++ b/tests/UriTest.php
@@ -113,17 +113,17 @@ class UriTest extends TestCase
     public function testPortMustBeValid()
     {
         $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('Invalid port: 100000. Must be between 1 and 65535');
+        $this->expectExceptionMessage('Invalid port: 100000. Must be between 0 and 65535');
 
         (new Uri())->withPort(100000);
     }
 
-    public function testWithPortCannotBeZero()
+    public function testWithPortCannotBeNegative()
     {
         $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('Invalid port: 0. Must be between 1 and 65535');
+        $this->expectExceptionMessage('Invalid port: -1. Must be between 0 and 65535');
 
-        (new Uri())->withPort(0);
+        (new Uri())->withPort(-1);
     }
 
     public function testParseUriPortCannotBeZero()
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -278,8 +278,8 @@ final class Uri implements UriInterface
         }
 
         $port = (int) $port;
-        if (1 > $port || 0xffff < $port) {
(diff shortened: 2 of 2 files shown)


Zegnat:
When do you need an HTTP message to be bound to port 0?

Clearly I do not know enough about networking to really say whether this is a good change or not. @Nyholm do you know why this was initially limited to port 1 and up?

Port 0 does exist on a theoretical level, in that the IANA reserves it and RFC 3986 (which PSR-7 uses for URIs) mentions it as part of “the "well-known port" range”. But AFAIK it is impossible to bind any listeners to on Unix systems, which makes it pretty useless. (Trying to bind to port 0 leads to binding to the next available port number.)

Nyholm:
Hm. If the RFC specify it as being valid and @tracerneo has an issue. Then I think we can agree with this change, right?

tests/UriTest.php
```diff
@@ -113,19 +113,11 @@ public function getInvalidUris()
...
 
         (new Uri())->withPort(100000);
     }
 
-    public function testWithPortCannotBeZero()
-    {
-        $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('Invalid port: 0. Must be between 1 and 65535');
-
-        (new Uri())->withPort(0);
```
Nyholm: Could you modify this to use -1 as port?

Tracerneo:
@Zegnat You are right TCP and UDP 0 ports cannot be bound, but you can listen on an [unix(7) socket file](https://linux.die.net/man/7/unix), which doesn't have port at all, and is reported by web server as nothing, which can turn into 0. Unix socket don't have sin_port field like [ip(7) sockets](https://linux.die.net/man/7/ip) do - compare sockaddr_in and sockaddr_un struct.

For php config would look like
```ini
[www]
listen = /var/run/php7-fpm.sock
```
And for [nginx](https://nginx.org/en/docs/http/ngx_http_core_module.html#listen):
```nginx
server {
  listen unix:/var/run/nginx.sock; # the bug triggering part
  location ~ \.php$ {
    fastcgi_pass unix:/var/run/php7-fpm.sock;
  }
}
```
Then you can connect to `/var/run/nginx.sock` using for example a load balancer.

@Nyholm Done.

tests/UriTest.php
Tracerneo: That's fixed now.

Zegnat:
> […] is reported by web server as nothing, which can turn into 0.

That almost feels like a bug to me. `http://192.0.2.0:0/` seems like an incorrect URI to me. Actually, since the socket connection wouldn’t be exposed, it feels doubly weird to me to include a port in URIs.

That said, I have no aversion to support a port 0 in this PSR-7 implementation. People who need 100% correct and verifiable URIs should probably be looking at something like [League\Uri](https://uri.thephpleague.com/) anyway.

Zegnat:
Rerequesting review. Looks to me like previous request for change has been fulfilled.

Issue #116, opened 2019-04-27
createResponse on factory creates missing reasonPhrase


dominikzogg:
PSR17's createResponse

```php
public function createResponse(int $code = 200, string $reasonPhrase = ''): ResponseInterface;
```
means, that there is never a null value on $reasonPhrase

this means that within the Response implementation, using the factory is always empty, cause the null check https://github.com/Nyholm/psr7/blob/master/src/Response.php#L49.

what is OK, but dangerous is also line https://github.com/Nyholm/psr7/blob/master/src/Response.php#L50 cause its checks against the property and the assignment agains the variable within the method.

dominikzogg:
I would suggest the following:

This would not only fix my issue, but also the issue given null as reason which a not supported status code, which leads to an interface implementation issue cause reasonPhrase should not return null as far as i unterstand the PSR7.

```php
// If we got no body, defer initialization of the stream until Response::getBody()
if ('' !== $body && null !== $body) {
    $this->stream = Stream::create($body);
}

$this->statusCode = $status;
$this->reasonPhrase = (string) $reason;

$this->setHeaders($headers);

if ('' === $this->reasonPhrase && isset(self::PHRASES[$this->statusCode])) {
    $this->reasonPhrase = self::PHRASES[$this->statusCode];
}

$this->protocol = $version;
```

PR #117, opened 2019-04-28
fix-response-reasonphrase-created-by-factory-and-type-safety

dominikzogg:
I could not write a test, which shows the type conversion issue, cause the tests is not using strict, (if i change that i break other tests), but hopefully you get the point within #116

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -45,12 +45,13 @@ final class Response implements ResponseInterface
         }
 
         $this->statusCode = $status;
-        $this->setHeaders($headers);
-        if (null === $reason && isset(self::PHRASES[$this->statusCode])) {
-            $this->reasonPhrase = self::PHRASES[$status];
-        } else {
-            $this->reasonPhrase = $reason;
+        $this->reasonPhrase = (string) $reason;
+
+        if ('' === $this->reasonPhrase && isset(self::PHRASES[$this->statusCode])) {
+            $this->reasonPhrase = self::PHRASES[$this->statusCode];
         }
+
+        $this->setHeaders($headers);
 
         $this->protocol = $version;
     }
@@ -77,11 +78,13 @@ final class Response implements ResponseInterface
         }
 
         $new = clone $this;
+
         $new->statusCode = $code;
-        if ((null === $reasonPhrase || '' === $reasonPhrase) && isset(self::PHRASES[$new->statusCode])) {
-            $reasonPhrase = self::PHRASES[$new->statusCode];
+        $new->reasonPhrase = (string) $reasonPhrase;
+
+        if ('' === $new->reasonPhrase && isset(self::PHRASES[$new->statusCode])) {
+            $new->reasonPhrase = self::PHRASES[$new->statusCode];
         }
-        $new->reasonPhrase = $reasonPhrase;
 
         return $new;
     }
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Nyholm\Psr7;
 
+use Nyholm\Psr7\Factory\Psr17Factory;
 use Nyholm\Psr7\Response;
 use PHPUnit\Framework\TestCase;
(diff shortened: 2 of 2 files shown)


dominikzogg:
#116

Zegnat:
Hmm. This looks like a breaking change, so whether we merge it depends on if we really want to treat this as a bug or not.

PSR-17 says we `MAY` use a default reason if none is given, but that is not required. It is a perfectly valid HTTP response without reason phrase. That’s why an empty string is treated differently from `null` by our constructor, there just isn’t a way to support the `null` value in the PSR-17 factories.

In fact, this PR disables empty reason phrases completely. Here is an example test that currently passes, but breaks after applying this PR:

```php
    public function testCanConstructWithStatusCodeAndEmptyReason()
    {
        $r = new Response(404, [], null, '1.1', '');
        $this->assertSame(404, $r->getStatusCode());
        $this->assertSame('', $r->getReasonPhrase());
    }
```

I am wondering if this isn’t a documentation error on our end, where we just need to point out that we do not implement that specific `MAY` section of PSR-17.

All of that said, you did uncover an actual bug (from #116):

> […] the issue given null as reason which a not supported status code, which leads to an interface implementation issue cause reasonPhrase should not return null as far as i unterstand the PSR7.

This can be shown by the following test:

```php
    public function testCanConstructWithNullReason()
    {
        $r = new Response(890, [], null, '1.1', null);
        $this->assertSame('', $r->getReasonPhrase());
    }
```

A PR that tests the different reason cases for the Response constructor and fixes them would definitely be accepted as it would not introduce breaking changes.

dominikzogg:
@Zegnat the reasonPhrase can be an empty string, but at the moment using strict types if leads to null, which breaks the PSR. Please check the withStatus there null and string is/was handled the same, so its at least inconsitent.

dominikzogg:
For those having the same "issue" (means those who used the slim and the zendframework ones before) using the factory, so no change on this library is needed.

PS: 

```
$responseFactory =  new class() implements ResponseFactoryInterface
{
    public function createResponse(int $code = 200, string $reasonPhrase = ''): ResponseInterface
    {
        return new Response($code, [], null, '1.1', '' !== $reasonPhrase ? $reasonPhrase : null);
    }
};
```

Zegnat:
> the reasonPhrase can be an empty string, but at the moment using strict types if leads to null, which breaks the PSR.

I am not sure what you mean with strict types leading to anything? Can you elaborate?

> Please check the withStatus there null and string is/was handled the same, so its at least inconsitent.

Good catch! I think this inconsistency definitely needs fixing. Though that too would be a backwards compatibility break. Looks like we put ourselves in a corner there.

> I cannot write this test fix the bug cause tests, are not written in strict mode

What test can’t you write? What scenario did you want to test? In my previous comment I included a test (`testCanConstructWithNullReason`) that shows we have a bug when passing in `null`. I’d love to see similar tests about the issues you are mentioning here so we can look into them.

As for your second comment, I would recommend you never depend on a PSR-17 factory to set a reason phrase. Because the standard does not require one to be set at all. If reason phrases are important to you, and you need the ability to switch between different PSR implementations, make sure to always define the reason phrase yourself when calling `createResponse()`.

dominikzogg:
@Zegnat 

```
public function testWillBreakIfTheTestClassIsUsingStrictTypesCauseItsNotAllowedToReturnNull()
{
    $r = new Response(700);
    $this->assertSame(700, $r->getStatusCode());
    $this->assertSame('', $r->getReasonPhrase());
}
```

Zegnat:
Yeah, that matches the test I wrote above. I am not sure strict types needs to be there for this to fail? It failed for me by default when I include it, which is why I was confused about your statement you couldn’t write the test.

That is definitely a bug that needs fixing and a PR for that would be accepted directly as it does not introduce any BC breaks. If you want to take a shot at that, go ahead. Else I’ll try to find some time during the upcoming week.

dominikzogg:
You're test was running without issues when I did the same

PR #118, opened 2019-05-09
Fix making default value of response phrase behavior to conform PSR-17

acgrid:
The factory may use default phrases in case of empty string according to https://github.com/php-fig/http-factory/blob/master/src/ResponseFactoryInterface.php#L17
Notice the default value of its second parameter in the method signature.

Test case appended towards `Nyholm\Psr7\Factory\Psr17Factory`, `testCreateResponse` will test both cases in which default and custom response reason provided.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -46,7 +46,7 @@ final class Response implements ResponseInterface
 
         $this->statusCode = $status;
         $this->setHeaders($headers);
-        if (null === $reason && isset(self::PHRASES[$this->statusCode])) {
+        if (($reason === null || $reason === '') && isset(self::PHRASES[$this->statusCode])) {
             $this->reasonPhrase = self::PHRASES[$status];
         } else {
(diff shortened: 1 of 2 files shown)


Zegnat:
This has been brought up previously in #117, but it is important to keep in mind that this is a breaking change. Currently we support people setting an empty reason phrase, with a change like this that will no longer work.

E.g. the following test works today but breaks with this PR:

```
    public function testCanConstructWithStatusCodeAndEmptyReason()
    {
        $r = new Response(404, [], null, '1.1', '');
        $this->assertSame(404, $r->getStatusCode());
        $this->assertSame('', $r->getReasonPhrase());
    }
```

Nyholm:
Thank you for this PR. I've done a slightly different implementation of it here #122 
This allows us to not break existing functionality in the Response class

Note: The PSR17Factory is tested with https://github.com/http-interop/http-factory-tests

PR #121, opened 2019-07-03
Fix checking for seekable stream resources

nicolas-grekas:
Userland stream wrappers always return true for the "seekable" metadata, yet this can be wrong. Here is a more accurate check.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -76,7 +76,7 @@ final class Stream implements StreamInterface
             $new = new self();
             $new->stream = $body;
             $meta = \stream_get_meta_data($new->stream);
-            $new->seekable = $meta['seekable'];
+            $new->seekable = $meta['seekable'] && 0 === \fseek($new->stream, 0, \SEEK_CUR);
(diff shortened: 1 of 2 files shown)


Zegnat:
This is a clever seek test. Looks great. Wondering why Travis is complaining here 🤔 I’ll try and take this for a test spin later this week and do an actual review.

hermann8u:
Hi,

Am I missing something or the parameters of the `fseek` function are not in the right order?

Should'nt it be like this:
`\fseek($new->stream, 0, \SEEK_CUR)`

nicolas-grekas:
>  \fseek($new->stream, 0, \SEEK_CUR)

indeed! fixed thanks.

Zegnat:
Can we add a test that breaks with the current code but is fixed by this PR? To make sure we do not regress this at a future point?
Is there a small enough “userland stream wrapper” that it makes sense to include one?

nicolas-grekas:
Test added.

PR #122, opened 2019-07-21
Added test to make sure we can create response with empty reason phrase

Nyholm:
Follow up form #118 

This PR also changes the behavior if you call `$psr17Factory->createResponse(200)`. 

Before we did not give any reason phrase, now we do. Is this a BC break? It is still fine according to PSR17.

### DIFF
--- a/src/Factory/Psr17Factory.php
+++ b/src/Factory/Psr17Factory.php
@@ -20,6 +20,11 @@ final class Psr17Factory implements RequestFactoryInterface, ResponseFactoryInte
 
     public function createResponse(int $code = 200, string $reasonPhrase = ''): ResponseInterface
     {
+        if (2 > \func_num_args()) {
+            // This will make the Response class to use a custom reasonPhrase
+            $reasonPhrase = null;
+        }
+
         return new Response($code, [], null, '1.1', $reasonPhrase);
     }
 
--- a/tests/Factory/Psr17FactoryTest.php
+++ b/tests/Factory/Psr17FactoryTest.php
@@ -20,5 +20,19 @@ class Psr17FactoryTest extends TestCase
 
         $r = $factory->createResponse(200, 'Foo');
         $this->assertEquals('Foo', $r->getReasonPhrase());
+
+        /*
+         * Test for non-standard response codes
+         */
+        $r = $factory->createResponse(567);
+        $this->assertEquals('', $r->getReasonPhrase());
+
+        $r = $factory->createResponse(567, '');
+        $this->assertEquals(567, $r->getStatusCode());
+        $this->assertEquals('', $r->getReasonPhrase());
+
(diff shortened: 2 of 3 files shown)


src/Factory/Psr17Factory.php
```diff
@@ -20,6 +20,11 @@ public function createRequest(string $method, $uri): RequestInterface
 
     public function createResponse(int $code = 200, string $reasonPhrase = ''): ResponseInterface
     {
+        if (2 > \func_num_args()) {
```
Zegnat: This is interesting, never thought of doing it this way! Basically a trick to emulate the method signature `string $reasonPhrase = null` without breaking the interface implementation?

tests/Factory/Psr17FactoryTest.php
```diff
@@ -0,0 +1,24 @@
...
+use Nyholm\Psr7\Factory\Psr17Factory;
+use PHPUnit\Framework\TestCase;
+
+class Psr17FactoryTest extends TestCase
+{
+    public function testCreateResponse()
+    {
+        $factory = new Psr17Factory();
+        $r = $factory->createResponse(200);
+        $this->assertEquals('OK', $r->getReasonPhrase());
```
Zegnat: Should this be extended to also include a test for a code that doesn’t have a predefined reason phrase? `567` or something otherwise weird? Or can we ignore that because we test that case in ResponseTest.php already? (`ResponseTest::testCanConstructWithUndefinedStatusCode`)

tests/ResponseTest.php
```diff
@@ -259,4 +259,11 @@ public function testHeaderValuesAreTrimmed($r)
         $this->assertSame('Foo', $r->getHeaderLine('OWS'));
         $this->assertSame(['Foo'], $r->getHeader('OWS'));
     }
+
+    public function testCanConstructWithStatusCodeAndEmptyReason()
```
Zegnat: This one looks good to me! I would maybe have preferred it at the top of the file together with `testCanConstructWithStatusCode` and `testCanConstructWithUndefinedStatusCode` but that is super nitpicky and more of a personal preference for grouping related tests.

src/Factory/Psr17Factory.php
Nyholm: Yeah, basically. 

I check to see how many arguments that were actually used.

tests/Factory/Psr17FactoryTest.php
Nyholm: Lets make the same test for the factory. 
Thank you

Nyholm:
Thank you for your review. I've updated according to feedback

Zegnat:
Updates look good to me!

Nyholm:
Thank you

PR #129, opened 2019-10-03
Fix to support UTF-8 paths

cseufert:


Zegnat:
This change seems to fail a number of the URL tests, including the one added by this PR? Are you still working on this?

I also wonder when you have had problems with `parse_url`? My local installation of PHP (7.3.7 on macOS) has no problem parsing `http://example.com/тестовый_путь/` with the default `parse_url` function.

cseufert:
I think your right, I will create a patch with a simpler fix for using strtolower() on hostnames with UTF-8 characters

PR #130, opened 2019-10-03
add php 7.4

andreybolonin:


Zegnat:
This looks good to me. Though a few questions (also aimed at @Nyholm):

1. Do we win anything by adding the snapshot now and upgrading to release version 7.4 within 2 months? Or are we just as well off waiting 2 months with adding thing?
2. When 7.4 is released, there is less than a week of planned time until PHP officially stops supporting 7.1. Do we drop 7.1 support at the same time PHP does, or do we keep legacy versions around for this library?
3. We already do PHP nightly testing, but that is described as PHP 8.0. Is there some way we can test both the PHP 8.0 and the PHP 7.X branch at all times? (Where X will be 5 once 7.4 is released?)

.travis.yml
```diff
@@ -9,6 +9,7 @@ php:
     - 7.1
     - 7.2
     - 7.3
+    - 7.4snapshot
```
RinkAttendant6: PHP 7.4 has been released:

```suggestion
    - 7.4
```

Nyholm:
Sorry for the delay. 

Thank you for this PR.

PR #133, opened 2019-10-22
Use a Unicode safe version of strtolower for hostnames

Zegnat:
This was a bit of a head-scratcher. After a long back-and-forth with @cseufert in #131 this is my try at fixing it. I am not sure this is better or worse than what is proposed in #131. It is definitely more complex, and hopefully covers the issue in its entirety for a multitude of platforms.

In the first commit I introduce a way to test for this problem in PHPUnit. I am not sure this is the best way to do it. I try to change the locale to something else if it happens to be `C` or `POSIX`. But there is no sure way to know what it should be changed to. Some form of `en_US` sounded like it would be most likely to be available.

I am not sure I like this way of testing. I cannot guarantee that the CI testing runs on has some form of `en_US` available, so we may still accidentally break the function without noticing. Doing setup and teardown within the test method itself is also bad, especially since the teardown (resetting the locale to what it originally was) fails if any of the PHPUnit assertions fail.

An alternative would be to split locale aware tests off to their own PHPUnit TestCase and use actual `setUp` and `tearDown` fixtures.

In the second commit I introduce a safe version of `strtolower`. It will either temporarily put the locale to `C` or `POSIX`, or use a regular expression only targetting ASCII characters. For this I assume that switching the locale to run `strtolower` only once is faster than running regular expressions and multiple `strtolower` calls.

The regular expression loop is hopefully slightly more performant than the one suggested by @cseufert as well. Instead of running for every single capital letter, it runs once per string of ASCII characters. A mixed string like `eXaMpLe.CoM` will only call `strtolower` once as the entire string will be matched in one go.

A problem with having the regular expression fallback included in the same method is that it complicates unit testing. On a system where `setlocale` works, it will never reach the regular expression variant. One solution is to mock `setlocale`, but then we must call the function unqualified (without `\`) or rely on a PHP extension to replace internals. If anyone has any good ideas, I am all ears!

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -54,7 +54,7 @@ final class Uri implements UriInterface
             // Apply parse_url parts to a URI.
             $this->scheme = isset($parts['scheme']) ? \strtolower($parts['scheme']) : '';
             $this->userInfo = $parts['user'] ?? '';
-            $this->host = isset($parts['host']) ? \strtolower($parts['host']) : '';
+            $this->host = isset($parts['host']) ? self::safeStrtolower($parts['host']) : '';
             $this->port = isset($parts['port']) ? $this->filterPort($parts['port']) : null;
             $this->path = isset($parts['path']) ? $this->filterPath($parts['path']) : '';
             $this->query = isset($parts['query']) ? $this->filterQueryAndFragment($parts['query']) : '';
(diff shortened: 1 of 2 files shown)


src/Uri.php
```diff
@@ -307,4 +307,26 @@ private static function rawurlencodeMatchZero(array $match): string
     {
         return \rawurlencode($match[0]);
     }
+
+    private static function safeStrtolower(string $string): string
```
nicolas-grekas: Same as https://github.com/Nyholm/psr7/pull/131/files#r359853348
` strtr($host, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz');` does the same but is much faster

Nyholm:
Thank you for this PR. 
I merge the solution in #143. I did however include this test (with your commit). 

Thank you again.

Issue #99, opened 2019-01-07
Inconsistent Stream state in factory methods (?)


mindplay-dk:
This issue may be a bit controversial, because the spec (PSR-17) appears to be incomplete on this point.

The issue is with the state of the internal stream produced by `createStreamFromFile()` and `createStream()` respectively.

While `createStreamFromFile()` produces a stream that points to the *start* of the file, `createStream()` produces a stream that points to the *end* of the file.

Producing a stream that points to the *end* of the file is problematic, since the most likely use of these streams is in the body of a `Response` - and you end up producing an empty response.

It's arguably a bit inconsistent and surprising, as these are just two different ways to create a stream from some pre-existing content - I think it's reasonable to expect them to work the same?

Diactoros is a popular source of reference for this, and it [does rewind the stream](https://github.com/zendframework/zend-diactoros/blob/master/src/StreamFactory.php#L30) when it creates a stream from a string.

Creating a stream from a string with the intent of writing to it is a much less likely scenario - if I *was* doing that, I'd probably start with `createStream("")` anyhow, and it would work either way, since the file pointer would be at 0 either way.

This has already lead to some pretty serious bugs while switching from Diactoros to this package.

Can we change it please?

Let me know if you'd accept a PR.

mindplay-dk:
Looks like there's already an ongoing discussion about this issue [here](https://groups.google.com/forum/#!topic/php-fig/S5YIw-Pu1yM).

Zegnat:
> Let me know if you'd accept a PR.

For now: no. (Unless @Nyholm over-rules me.) There was a PR for this already, #98, but it was closed by its author after I blocked it pending further discussion. That said:

I am super aware of this issue, as I am the person who also raised it (and did the multi-implementation testing) over on the mailing list. This will definitely be fixed in some way, but before diving in head first I would like to give the PHP-FIG WG a chance to discuss. This will hopefully ensure we do not end up implementing something today that needs to be redone tomorrow.

For now, if you need the ability to switch between PSR-7 implementations, you have two options:

1. Cast the object to a string, `__toString()` is specified by PHP-FIG to return as much content as possible (from start to end, if seeking to the beginning is possible on the current resource).
2. Always use some combination of `tell()`/`seek()`/`rewind()` if you depend on using `read()`.

Nyholm:
I agree with @Zegnat. 

I see no reason to change unless a PSR is updated with clear guidelines. 

[...]

PR #138, opened 2019-12-17
Fix bug with stream rewind in Stream::create

vadimzzz:
Your stream is left unrewinded after creation and it breaks compatibility with other PSR-7 libraries because they does not require manual stream rewind.
This looks like a bug, so I propose a fix.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -76,7 +76,7 @@ final class Stream implements StreamInterface
             $new = new self();
             $new->stream = $body;
             $meta = \stream_get_meta_data($new->stream);
-            $new->seekable = $meta['seekable'] && 0 === \fseek($new->stream, 0, \SEEK_CUR);
+            $new->seekable = $meta['seekable'] && 0 === \fseek($new->stream, 0, \SEEK_SET);
             $new->readable = isset(self::READ_WRITE_HASH['read'][$meta['mode']]);
             $new->writable = isset(self::READ_WRITE_HASH['write'][$meta['mode']]);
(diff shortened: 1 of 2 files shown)


Zegnat:
This feels like it is related to #99, we wanted to wait for an actual decision from PHP-FIG’s end before we change to a different behaviour.

Let me know if I am reading your PR wrong, but seems related. PHP-FIG does not specify where the cursor should go, so all. of the different conflicting implementations are actually valid.

vadimzzz:
Yes it seems related. I think in case of no standard for this it's better to implement the same way as other popular libraries do (btw it's obvious way for me), just because it may save some minutes of debugging trying to figure out why request body is emty.

nicolas-grekas:
This change is wrong: it changes the state of the stream, while all we want here is computing a boolean.

Nyholm:
Thank you for this PR. 

I think this should be closed for the following reasons: 
- PSR7 does not specify where the cursor of the stream should be
- We don't want to rewind the stream if we don't have to
- If a middleware or implementation assume the position of the cursor, they are wrong*. 

*Yes, saying "they are wrong" is harsh. But what if you had a middleware that move the cursor of the stream to a random position. If you included that middleware in your stack, then all your application would break... 

I want to make a point that if you have a middleware (or any consumer of a stream) that wants to read a stream, that middleware has to make sure it is reading the full stream, not just "the rest" of the stream. 

A super simple way of doing this is to use the `__toString()` function. 

--------

I will be happy to change this if PSR7 is updated to include the position of the cursor in the stream.

PR #144, opened 2020-04-20
Make __toString() compatible with throwing

### LINKED ISSUE #142: Response::__toString() should not hide exception
https://github.com/Nyholm/psr7/blob/d38874cbc289d25a309cefd04379e24352ced608/src/Stream.php#L98-L109

I found this because of https://github.com/php-http/client-common/issues/186
In my application, I had a 200 with an empty body :exploding_head: 

IMHO, doing this is wrong as it hides the root cause.

More over, in PHP 7.4 it's allowed to throw exception.
And for PHP 7.3-, it prefer a fatal error than something wrong.

WDYT?

nicolas-grekas:
Fix #142

On PHP 7.4, `__toString()` can throw.
And before PHP  7.4, `__toString()` can throw also, when the Symfony ErrorHandler is used: https://github.com/symfony/symfony/blob/e46495499841f805ab1915643f763d736d24822e/src/Symfony/Component/ErrorHandler/ErrorHandler.php#L498-L503

This fixes tests meanwhile.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Nyholm\Psr7;
 
 use Psr\Http\Message\StreamInterface;
+use Symfony\Component\Debug\ErrorHandler as SymfonyLegacyErrorHandler;
+use Symfony\Component\ErrorHandler\ErrorHandler as SymfonyErrorHandler;
 
 /**
  * @author Michael Dowling and contributors to guzzlehttp/psr7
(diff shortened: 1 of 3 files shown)


src/Stream.php
```diff
@@ -95,15 +97,31 @@ public function __destruct()
         $this->close();
     }
 
-    public function __toString(): string
+    /**
+     * @return string
```
nicolas-grekas, line 101: this is required to allow Symfony's error handler to work effectively.

Nyholm:
Do you mind rebasing this PR?

nicolas-grekas:
Thank you for merging!

PR #147, opened 2020-04-21
Prepare 1.3.0

Nyholm:
Change log for 1.3.0. After this is merged I want to tag a release. 

Diff: https://github.com/Nyholm/psr7/compare/1.2.1...master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## 1.3.0
+
+### Added
+
+- Make Stream::__toString() compatible with throwing exceptions on PHP 7.4.
+
+### Fixed
+
+- Support for UTF-8 hostnames
+
 ## 1.2.1
 
 ### Changed
@@ -13,9 +23,9 @@ All notable changes to this project will be documented in this file, in reverse 
 ### Changed
 
 - Change minimal port number to 0 (unix socket)
-- Updated `Psr17Factory::createResponse` to respect the specification. If second 
-  argument is not used, a standard reason phrase. If an empty string is passed, 
-  then the reason phrase will be empty. 
+- Updated `Psr17Factory::createResponse` to respect the specification. If second
+  argument is not used, a standard reason phrase. If an empty string is passed,
+  then the reason phrase will be empty.
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


Zegnat:
I have also been staring at that Roave error for a while now, and am still not quiet sure why it thinks we are updating any of the test packages. Can someone tell me? 🤔

Nyholm:
Sorry for leaving this hanging. 

About the Roave error: 
Our last minor version (1.2.0) had a dev requirement on `php-http/psr7-integration-tests: dev-master`. Since that package has updated to only support PHPUnit8 (and we only support PHPUnit7). That means that we cannot install nyholm/psr7:1.2.0 with dev dependencies anymore.

This will be fixed in 1.3.0 since we require a fixed version of php-http/psr7-integration-tests instead of dev-master.

Nyholm:
Considering this StreamInterface: 

```php
interface StreamInterface
{
    /**
     * Reads all data from the stream into a string, from the beginning to end.
     *
     * This method MUST attempt to seek to the beginning of the stream before
     * reading data and read the stream until the end is reached.
     *
     * Warning: This could attempt to load a large amount of data into memory.
     *
     * This method MUST NOT raise an exception in order to conform with PHP's
     * string casting operations.
     *
     * @see http://php.net/manual/en/language.oop5.magic.php#object.tostring
     * @return string
     */
    public function __toString();
```

As argued in #144, The following sentence: 
> This method MUST NOT raise an exception in order to conform with PHP's string casting operations.

Is intrepid as: 

> Because of PHP does not allow you to throw exceptions in __toString(), this method cannot throw exceptions either. 

But because PHP7.4 can throw exceptions, the second half in invalid. Ie, no BC break. 

As also discussed in #142, hiding exceptions it not something anybody appreciates. It just causes long debugging sessions.

Zegnat:
> But because PHP7.4 can throw exceptions, the second half in invalid. Ie, no BC break.

I don’t think it is not a BC break. Previously it was guaranteed that casting to a string would not throw, and after this upgrade it may start throwing for people. This means their application logic could break and they would have to adapt by adding try-catch blocks.

I do think this is a minor thing. And honestly, if we were surpressing exceptions in their code previously it may be good for them to get to the core issue of the exception. It is just something I would want to call attention too 😄 

Like I said before, happy to not fight on just that and instead get a new tag in people’s hands!

Nyholm:
>  Previously it was guaranteed that casting to a string would not throw, and after this upgrade it may start throwing for people.

Yes. But that is the same with all `__toString()` functions. The BC break comes from PHP 7.4.

Nyholm:
Thank you

PR #149, opened 2020-04-23
Added support for numeric header names

nickdnk:
Hello

As discussed at https://github.com/Nyholm/psr7-server/issues/35, here are some changes that seem to solve the problem.

This may be a premature PR, but it was the easiest way for me to demonstrate a cohesive solution with tests that works. @Zegnat did most of this, I just put it together.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -140,6 +140,11 @@ trait MessageTrait
     private function setHeaders(array $headers): void
     {
         foreach ($headers as $header => $value) {
+            if (\is_int($header)) {
+                // If a header name was set to a numeric string, PHP will cast the key to an int.
+                // We must cast it back to a string in order to comply with validation.
+                $header = (string) $header;
+            }
             $value = $this->validateAndTrimHeader($header, $value);
             $normalized = self::lowercase($header);
             if (isset($this->headerNames[$normalized])) {
--- a/tests/RequestTest.php
+++ b/tests/RequestTest.php
@@ -222,6 +222,9 @@ class RequestTest extends TestCase
         $this->assertSame(['NumericHeaderValue'], $r->getHeader('200'));
         $this->assertSame('NumericHeaderValue', $r->getHeaderLine('200'));
 
+        $this->assertSame(['NumericHeaderValueZero'], $r->getHeader('0'));
+        $this->assertSame('NumericHeaderValueZero', $r->getHeaderLine('0'));
+
         $r = $r->withHeader('300', 'NumericHeaderValue2')
                ->withAddedHeader('200', ['A', 'B']);
 
--- a/tests/RequestTest.php
+++ b/tests/RequestTest.php
@@ -202,6 +202,48 @@ class RequestTest extends TestCase
         $this->assertSame('200', $r->getHeaderLine('Content-Length'));
     }
 
+    public function testSupportNumericHeaderNames()
+    {
+        $r = new Request(
+            'GET', '', [
+                     '200' => 'NumericHeaderValue',
+                     '0'   => 'NumericHeaderValueZero',
+                 ]
+        );
+
+        $this->assertSame(
+            [
+                '200' => ['NumericHeaderValue'],
+                '0'   => ['NumericHeaderValueZero'],
+            ],
+            $r->getHeaders()
+        );
+
+        $this->assertSame(['NumericHeaderValue'], $r->getHeader('200'));
+        $this->assertSame('NumericHeaderValue', $r->getHeaderLine('200'));
+
+        $r = $r->withHeader('300', 'NumericHeaderValue2')
+               ->withAddedHeader('200', ['A', 'B']);
+
+        $this->assertSame(
+            [
+                '200' => ['NumericHeaderValue', 'A', 'B'],
+                '0'   => ['NumericHeaderValueZero'],
+                '300' => ['NumericHeaderValue2'],
+            ],
+            $r->getHeaders()
+        );
+
+        $r = $r->withoutHeader('300');
+        $this->assertSame(
+            [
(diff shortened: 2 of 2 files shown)


src/MessageTrait.php
```diff
@@ -140,6 +140,7 @@ public function withBody(StreamInterface $body): self
     private function setHeaders(array $headers): void
     {
         foreach ($headers as $header => $value) {
+            if (is_int($header)) $header = strval($header);
```
Zegnat, line 143: This one-liner needs to be changed to actually match the project’s style guide. Running `php-cs-fixer` gives us:

```suggestion
            if (\is_int($header)) {
                $header = (string) $header;
            }
```

src/MessageTrait.php
```diff
@@ -92,7 +92,7 @@ public function withHeader($header, $value): self
 
     public function withAddedHeader($header, $value): self
     {
-        if (!\is_string($header) || '' === $header) {
+        if (!\is_int($header) && (!\is_string($header) || '' === $header)) {
```
Zegnat, line 95: [PHPStan is complaining](https://github.com/Nyholm/psr7/pull/149/checks?check_run_id=613771113) about this addition. And I think rightfully so. Removing this `is_int()` call does not break any tests. If we need this here, could you add a test case that requires it?

[PSR-7](https://www.php-fig.org/psr/psr-7/) only allows string values for `$header` here, unless I am [misreading something](https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/MessageInterface.php#L145-L150):

```php
    /**
     * […snip…]
     * @param string $name Case-insensitive header field name to add.
     * @param string|string[] $value Header value(s).
     * @return static
     * @throws \InvalidArgumentException for invalid header names or values.
     */
    public function withAddedHeader($name, $value);
```

So we shouldn’t need to account for integers here. Any application calling `MessageInterface::withAddedHeader` (or `MessageInterface::withHeader` for that matter) MUST provide strings.

tests/RequestTest.php
```diff
@@ -202,6 +202,38 @@ public function testSupportNumericHeaders()
...
+        $this->assertSame(['NumericHeaderValue'], $r->getHeader('200'));
+        $this->assertSame('NumericHeaderValue', $r->getHeaderLine('200'));
+
+        $r = $r->withHeader('300', 'NumericHeaderValue2')
+               ->withAddedHeader('200', ['A', 'B']);
+
+        $this->assertSame(
+            [
+                '200' => ['NumericHeaderValue', 'A', 'B'],
+                '300' => ['NumericHeaderValue2']
```
Zegnat, line 223:
```suggestion
                '300' => ['NumericHeaderValue2'],
```

tests/RequestTest.php
```diff
@@ -202,6 +202,38 @@ public function testSupportNumericHeaders()
...
+            ],
+            $r->getHeaders()
+        );
+
+        $r = $r->withoutHeader('300');
+        $this->assertSame(
+            [
+                '200' => ['NumericHeaderValue', 'A', 'B']
```
Zegnat, line 231:
```suggestion
                '200' => ['NumericHeaderValue', 'A', 'B'],
```

src/MessageTrait.php
nickdnk: I think this may need to be fixed [here](https://github.com/Nyholm/psr7-server/blob/e6a526e9170e6e33a13efc2b61703ca476b7ea68/src/ServerRequestCreator.php#L68) instead then. Since that's the function that passes integers to this method.

Zegnat: Correct! As Nyholm/psr7-server can work with any PSR-7 implementation and not just this one it should follow the spec and make sure to only pass in string values there.

nickdnk: Okay. See https://github.com/Nyholm/psr7-server/pull/36. That PR depends on this one to be merged.

Nyholm:
Thank you. 

If this library (or the PSR specification) was PHP 7.2+. It would have a type hint for `string` on `withAddedHeader`. In fact, the PSR specify that the `$header` should be a string. 

I dont think it is a good idea to convert an integer to a string. I think it is the callee (nyholm/psr7-server in this case) that should make sure it si giving a string to the `withAddedHeader` function.

Nyholm:
Hm.. I see. So if we do `->withAddedHeader('0', 'foo')`, we will convert that to an int because it is used as an array key.. hm.

src/MessageTrait.php
```diff
@@ -140,6 +140,9 @@ public function withBody(StreamInterface $body): self
     private function setHeaders(array $headers): void
     {
         foreach ($headers as $header => $value) {
+            if (\is_int($header)) {
```
Nyholm, line 143:
```suggestion
            // The array key could have been converted to an int. 
            if (\is_int($header)) {
```

nickdnk: Squashed commits and added a comment here.

nickdnk:
Squashed.

nickdnk:
Hey guys

Can we merge this in and update the dev-dependencies of https://github.com/Nyholm/psr7-server/pull/36/files? The tests over there require this component to be updated as psr7-server uses it for tests which fail on that PR because of this.

Nyholm:
Yes, This should be merged. Thank you

nickdnk:
I just added another test of `getHeader()` and `getHeaderLine()` with a value of `0`, just to future-proof this edge-case.

nickdnk:
Hey guys

Can we get this released so we can complete https://github.com/Nyholm/psr7-server/pull/36? I cannot fix the errors I get in my API without this, unless I switch to a different PSR7 implementation which I would like to avoid having to toy with at the moment.

@Nyholm @Zegnat

Issue #153, opened 2020-09-22
Return value of Nyholm\Psr7\Stream::read() must be of the type string, bool returned


casahugo:
In `Nyholm\Psr7\Stream`, the method `read` must return only string but `fread` can be return a boolean
The StreamInterface does not force return type
```php
    public function read($length): string
    {
        if (!$this->readable) {
            throw new \RuntimeException('Cannot read from non-readable stream');
        }

        return \fread($this->stream, $length);
    }
```

Zegnat:
Woops, silly mistake, well spotted!

Not sure exactly when it would return `false`. If we previously established the stream to be readable (as it passes the `if`) but no longer is readable, maybe? Trying to think how we could make a test case for this one.

Zegnat:
Fix is just a clone of how we handle this for `fwrite`:

```diff
diff --git a/src/Stream.php b/src/Stream.php
index 0dd8a2b..2b62e1c 100644
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -240,7 +240,11 @@ final class Stream implements StreamInterface
             throw new \RuntimeException('Cannot read from non-readable stream');
         }
 
-        return \fread($this->stream, $length);
+        if (false === $result = \fread($this->stream, $length)) {
+            throw new \RuntimeException('Unable to read from stream');
+        }
+
+        return $result;
     }
 
     public function getContents(): string
```

Which, I just realised, I am not sure we have tests for either. And it may be impossible to test for. Sounds like something substantial in the PHP/C level needs to go wrong for this to occur?

mahagr:
Maybe you should try to read negative amount of bytes?

casahugo:
Thank you very much for your response
I arrived at the same conclusions, the test seems difficult to write but I may have a track by deleting the resource during reading

mahagr:
Deleting the file doesn't work as unix filesystems keeps the inode reserved/open as long as you have handle on it. I actually have used it to my benefit -- until I noticed that it fails in Windows :(

Zegnat:
Let us bring the discussion about the actual test code into the open PR. Thanks for the quick PR, @casahugo! Keeping the bug open until we bring the PR to a conclusion.

PR #154, opened 2020-09-23
fix throw exception on read stream

casahugo:
fix https://github.com/Nyholm/psr7/issues/153

### DIFF
--- a/tests/StreamTest.php
+++ b/tests/StreamTest.php
@@ -66,15 +66,6 @@ class StreamTest extends TestCase
         $stream->close();
     }
 
-    public function testReadError()
-    {
-        $this->expectException(\RuntimeException::class);
-        $handle = fopen('php://temp', 'a');
-        $stream = Stream::create($handle);
-
-        $stream->read(4);
-    }
-
     public function testGetSize()
     {
         $size = filesize(__FILE__);
--- a/tests/StreamTest.php
+++ b/tests/StreamTest.php
@@ -66,6 +66,15 @@ class StreamTest extends TestCase
         $stream->close();
     }
 
+    public function testReadError()
+    {
+        $this->expectException(\RuntimeException::class);
+        $handle = fopen('php://temp', 'a');
+        $stream = Stream::create($handle);
+
+        $stream->read(4);
+    }
+
     public function testGetSize()
     {
         $size = filesize(__FILE__);
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -240,7 +240,11 @@ final class Stream implements StreamInterface
             throw new \RuntimeException('Cannot read from non-readable stream');
         }
 
-        return \fread($this->stream, $length);
+        if (false === $result = \fread($this->stream, $length)) {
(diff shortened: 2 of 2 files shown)


tests/StreamTest.php
```diff
@@ -66,6 +66,15 @@ public function testChecksEof()
         $stream->close();
     }
 
+    public function testReadError()
+    {
+        $this->expectException(\RuntimeException::class);
+        $handle = fopen('php://temp', 'a');
```
mahagr, line 72: Well, that does work. You cannot read from the write-only stream. :)

casahugo: Yes it's seem work, but why the CI throw an error on BC Check ?

Zegnat:
The newly added test case here does not test the previously wrong behaviour. Note that the following passes for append-only streams:

```php
        $handle = fopen('php://temp', 'a');
        $stream = Stream::create($handle);
        $this->assertFalse($stream->isReadable()); // ← important bit
```

This means the `RuntimeException` you are testing for in the new `testReadError` is actually the existing “Cannot read from non-readable stream” exception and not the newly added “Unable to read from stream”.

The new test case should probably be as follows, checking for the exact error:

```php
    public function testReadError()
    {
        $this->expectException(\RuntimeException::class);
        $this->expectExceptionMessage('Unable to read from stream');
        $handle = fopen('php://temp', 'a');
        $stream = Stream::create($handle);

        $stream->read(4);
    }
```

This will not pass, because the code there is not enough to trigger a return value of `false` from the `fread` :( 

So I am 👍 on the fix, which is the same [as I mentioned](https://github.com/Nyholm/psr7/issues/153#issuecomment-696855655), but the test case remains a head scratcher.

tests/StreamTest.php
```diff
@@ -66,6 +66,15 @@ public function testChecksEof()
         $stream->close();
     }
 
+    public function testReadError()
+    {
+        $this->expectException(\RuntimeException::class);
```
Zegnat, line 71: Use the following assertion to make sure you are testing the `RuntimeException` that is being introduced by this PR:
```suggestion
        $this->expectException(\RuntimeException::class);
        $this->expectExceptionMessage('Unable to read from stream');
```
For me this sadly breaks the test 😞 So doing a `->read(4)` from an append-only handle is not enough to trigger the `false` value.

Zegnat (changes requested):
Lets either fix the test or remove it from this PR.

tests/StreamTest.php
casahugo: I have no idea so that fread return false :disappointed:, just remove this test.

Zegnat: Thanks for the update!

Zegnat (approved):

Zegnat:
All looks good to me 🚀 

I just need to find a minute of extra time to see why the BC checks are failing before hitting merge. Thanks for the quick turnover @casahugo!

Nyholm (approved):
Thank you for the work and reviews.

I think it all looks good. The BC check is unrelated, see #155

PR #160, opened 2020-11-14
Add Github action to automatically update branch alias

Nyholm:
Just to make sure we never forget =)


Nyholm:
I've added similar action on async-aws/aws and this exact action on FakerPHP/Faker

PR #164, opened 2020-12-24
Use static return annotation for best IDE method suggestion

samizdam:


samizdam:
Also I prepare CS fixes and can merge its in forked master, if it is required:

https://github.com/samizdam/psr7/pull/1/files

Zegnat:
Static return types [are different from self](https://wiki.php.net/rfc/static_return_type) on the language level. And both are valid return types starting in PHP 8. So I would prefer to not put something in the DocBlock when it is not what is actually being returned.

Sounds like this should be a feature request to the IDE to add support for both static and self?

Am I missing something obvious?

PR #165, opened 2020-12-28
Rename Zend to Laminas

Groruk:
Small PR to rename Zend to Laminas and update package & namespace references to the newer Laminas releases, after Zend became the Laminas project in January 2020.

(See https://framework.zend.com/blog/2019-04-17-announcing-laminas.html for further reference)


Nyholm:
Thank you

PR #166, opened 2021-01-04
Updated Zend to Laminas in README

Shikachuu:
Since Zend is no longer maintained, the README should be updated.
- [X] Changed Zend to Laminas in test cases. (*)
- [X] Changed Zend HttpHandlerRunner to the Laminas one.

>  *: Create a Pull Request to the [HTTP Client Benchmark also](https://github.com/Nyholm/http-client-benchmark), because it is probably outdated and isn't 100% accurate for the 'new' Laminas case.
> This Pull Request probably the same as #165, but I'll make a follow up in the above mentioned repository.
> Edit: the follow up in the benchmarks also ready to merge: [Test PSR-7 implementation #2](https://github.com/Nyholm/http-client-benchmark/pull/2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ A super lightweight PSR-7 implementation. Very strict and very fast.
 
 | Description | Guzzle | Laminas | Slim | Nyholm |
 | ---- | ------ | ---- | ---- | ------ |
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this PR.

This is updated in master since a while back.

Issue #139, opened 2019-12-19
Why final?


leemason:
I understand the goal of the project is to be compliant with the psr spec, but i dont see why that would enforce your implementations to be final?

Being spec compliant, nothing less and nothing more is perfect for use as a parent class in projects where you want to extend the functionality to include your own little helpers here and there and not have to write your own implementation, or deal with the baggage of implementations that add their own methods outside of the spec.

Is this something you would consider changing?

mahagr:
You should use decorators instead of extending the classes:

https://github.com/getgrav/grav/tree/develop/system/src/Grav/Framework/Psr7/Traits
https://github.com/getgrav/grav/tree/develop/system/src/Grav/Framework/Psr7 (ignore AbstractUri, it's deprecated)

XedinUnknown:
Thing is, you can't always use decoration to change behaviour in a meaningful way, without re-implementing. In a perfect world, that would be great. But we don't write perfect code, and therefore I feel it is unfair to prevent users from adapting code to their needs. Also, and this is my opinion, making things explicitly `final` violates [OCP][].

[OCP]: https://en.wikipedia.org/wiki/Open/closed_principle

Nyholm:
These are excellent questions. This has been discussed in lengths at #31. 

I know that this library is forcing people to write good code and not to be lazy. I know that it is frustrating. I have not yet seen an argument based on computer science that show that the classes should not be final. 

> Also, and this is my opinion, making things explicitly final violates OCP.

I see that this is your opinion. But that opinion has no support in the definition of polymorphic open-closed principle. =) 

You have to consider the "module" to be the PSR-7 specification together with this implementation.

PR #169, opened 2021-02-10
Remove final keyword

Nyholm:
I don't think there is any benefit to remove the `final` keyword. All the good software scientists and best practises says that you should never efter extend these value objects. 

However, some people insists on writing code that I don't approve of and from their perspective the `final` keyword is just annoying. 

This PR does not change the concept, but technically it allows some people to write the code they want to write even tough I consider it suboptimal. 

-----

Edit: When I read this description again, it looks that Im condescending. Im really not. Im trying to be pragmatic.

### DIFF
--- a/doc/final.md
+++ b/doc/final.md
@@ -1,0 +1,20 @@
+# Final classes
+
+The `final` keyword was removed in version 1.4.0. It was replaced by `@final` annotation.
+This was done due popular demand, not because it is a good technical reason to
+extend the classes.
+
+This document will show the correct way to work with PSR-7 classes. The "correct way"
+refers to best practices and good software design. I strongly believe that one should
+be aware of how a problem *should* be solved, however, it is not needed to always
+implement that solution.
+
+## Extending classes
+
+You should never extend the classes, you should rather use composition or implement
+the interface yourself.
+
+## Mocking classes
+
+The PSR-7 classes are all value objects and they can be used without mocking. If
+one really needs to create a special scenario, one cna mock the interface instead.
--- a/doc/final.md
+++ b/doc/final.md
@@ -17,4 +17,4 @@ the interface yourself.
 ## Mocking classes
 
 The PSR-7 classes are all value objects and they can be used without mocking. If
-one really needs to create a special scenario, one cna mock the interface instead.
+one really needs to create a special scenario, one can mock the interface instead.
--- a/doc/final.md
+++ b/doc/final.md
@@ -12,7 +12,7 @@ implement that solution.
 ## Extending classes
 
 You should never extend the classes, you should rather use composition or implement
-the interface yourself.
+the interface yourself. Please refer to the [decorator pattern](https://refactoring.guru/design-patterns/decorator).
 
 ## Mocking classes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## 1.4.0
+
+### Removed
+
+The `final` keyword was replaced by `@final` annotation.
+
 ## 1.3.2
 
 ### Fixed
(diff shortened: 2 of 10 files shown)


Zegnat:
I also really dislike this, because extending a PSR implementation tends to people relying on the extended version, which in turn means they have just limited themselves to never be able to switch implementations. A problem that does not happen if you follow a decorator pattern. The entire point is interoperability. I almost feel this is just helping someone else get into trouble down the line.

I have been thinking about ways to help people build decorators, and wonder if it might be better to offer some sort of abstract class to extend that implements all the methods. E.g. we'd be providing the abstract class called `ADecoratorBase` from this example:

```php
interface AnInterface { // Imagine this is a PSR-7 value object interface.
    public function action(): bool;
}

final class AnImplementation implements AnInterface { // Imagine this is our implementation of said interface.
    public function action(): bool {
        return true;
    }
}

abstract class ADecoratorBase implements AnInterface { // Imagine this is a "decorator" we provide.
    protected AnInterface $base;

    public function action(): bool {
        return $this->base->action();
    }
}

class MyDecorator extends ADecoratorBase { // Now all a user needs to do to decorate is this.
    public function __construct(AnInterface $base)
    {
        $this->base = $base;
    }
}

$obj = new MyDecorator(new AnImplementation());
assert($obj->action() === true, 'Accessing expected method from base.');
```

Alternatively shipping something like the Grav traits that @mahagr linked to in https://github.com/Nyholm/psr7/issues/139#issuecomment-567828427.

Just wondering if we can do some sort of educative good by doing that, rather than just dropping the final keyword that we both believe as correct for this library :thinking:

mahagr:
I have to say that extending class also sounds tempting as it would allow customization without too much work. This is also where the traits come -- it will save a lot of work. The reason why I ended up using traits was to be able to inject the PSR-7 interface into any (existing) class.

I think it's also a good idea to separate decorator/traits from this library in order to prevent fixating code into a single library.

Nyholm:
@Zegnat, Yeah, but looking at the implementations, most methods just have a few lines, I don't think it make much sense to also add a bunch of abstract classes. 

What if we add a page in the documentation to better explain why extending classes are a bad idea. We can also show what one should do instead. The `@final` comment can link to that page.

Nyholm:
The PR is updated

doc/final.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Extending classes
+
+You should never extend the classes, you should rather use composition or implement
+the interface yourself.
+
+## Mocking classes
+
+The PSR-7 classes are all value objects and they can be used without mocking. If
+one really needs to create a special scenario, one cna mock the interface instead.
```
Zegnat, line 20:
```suggestion
one really needs to create a special scenario, one can mock the interface instead.
```

Zegnat (changes requested):
> Yeah, but looking at the implementations, most methods just have a few lines, I don't think it make much sense to also add a bunch of abstract classes.

Yet apparently people do not want to copy the few lines and make their own class. I also wouldn’t put the actual implementation in the abstract classes, only make an abstract decorator (wrapper) class that calls the methods from the actual wrapped value object.

But the more I think about that, the less I would want that as part of the Psr7 library. I might just create a Psr7Decorator package later…

> What if we add a page in the documentation to better explain why extending classes are a bad idea. We can also show what one should do instead. The `@final` comment can link to that page.

I like adding implicit the docs.

I wonder if we should point to some patterns that people might actually be looking for, rather than extending. Such as [decorating](https://refactoring.guru/design-patterns/decorator). There is nothing wrong with wanting to wrap a value object adding additional logic to match the application.

Nyholm:
Thank you.

Let me improve the docs with some more references.

frodeborli:
First of all; I like the quality of your work - which is why I decided to use your PSR-7 implementation. Unfortunately, I have to drop it or fork it due to the `@final` keyword.

The PSR does not define the signature for the `__construct` method. Since your class prohibits overriding the constructor - thanks to the `@final` keyword, I don't think your implementation complies 

I'm sad to learn that you have decided to mark your class with `@final`, for *political* reasons. Reason seems to be that you think I should change my style of programming because you think composition is awesome...

PR #143, opened 2020-04-20
Add locale-independent lowercasing

nicolas-grekas:
Fix #131 and fix #133

PR #172, opened 2021-02-18
Remove LowercaseTrait

Nyholm:
This is not really needed. The tests pass anyways.

This was added in #143


src/Factory/Psr17Factory.php
```diff
@@ -52,7 +52,7 @@ public function createStreamFromResource($resource): StreamInterface
         return Stream::create($resource);
     }
 
-    public function createUploadedFile(StreamInterface $stream, int $size = null, int $error = \UPLOAD_ERR_OK, string $clientFilename = null, string $clientMediaType = null): UploadedFileInterface
+    public function createUploadedFile(StreamInterface $stream, int $size = null, int $error = UPLOAD_ERR_OK, string $clientFilename = null, string $clientMediaType = null): UploadedFileInterface
```
nicolas-grekas, line 55: maybe these cs changes are not needed?

nicolas-grekas (approved):

src/Factory/Psr17Factory.php
Nyholm: Very true.. php-cs-fixer suggested it, I just blindly approved.

Nyholm: Rules updated!

PR #174, opened 2021-03-08
Wrap fopen for PHP 8

Zegnat:
`fopen()` has new behaviour where it will throw an error in PHP 8. Some projects have bumped into this, e.g. [Drupal](https://www.drupal.org/project/drupal/issues/3177541) and [the HTTP Factory tests](https://github.com/http-interop/http-factory-tests/pull/46).

This PR wraps our uses of `fopen()` to not leak the `ValueError` to library users.

Marking as WIP as I haven’t looked at tests yet. But have a look and see if these changes seem right!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
-## Unreleased
+## 1.4.1
 
 ### Fixed
 
 - `Stream::create` with a string needs to rewind the created memory stream.
+- `Psr17Factory::createStreamFromFile`, `UploadedFile::moveTo`, and
+  `UploadedFile::getStream` no longer throw `ValueError` in PHP 8.
 
 ## 1.4.0
 
@@ -118,4 +120,3 @@ The `final` keyword was replaced by `@final` annotation.
 ## 0.2.3
 
 No changelog before this release
-
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7.5 || 8.5 || 9.4",
         "php-http/psr7-integration-tests": "^1.0",
-        "http-interop/http-factory-tests": "^0.8",
+        "http-interop/http-factory-tests": "^0.9",
         "symfony/error-handler": "^4.4"
     },
     "provide": {
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -61,6 +61,7 @@ class Stream implements StreamInterface
      * @param string|resource|StreamInterface $body
      *
      * @throws \InvalidArgumentException
+     * @throws \ValueError
      */
     public static function create($body = ''): StreamInterface
     {
@@ -87,6 +88,29 @@ class Stream implements StreamInterface
         }
 
         throw new \InvalidArgumentException('First argument to Stream::create() must be a string, resource or StreamInterface.');
+    }
+
+    /**
+     * Creates a new PSR-7 stream from path.
+     *
+     * @internal Implements Psr\Http\Message\StreamFactoryInterface::createStreamFromFile for internal use, library users should default to the actual Factory object
+     */
+    public static function createFromFile(string $filename, string $mode = 'r'): StreamInterface
+    {
+        try {
+            $resource = @\fopen($filename, $mode);
+        } catch (\Throwable $e) {
+            throw new \RuntimeException('The file ' . $filename . ' cannot be opened.');
+        }
+        if (false === $resource) {
+            if ('' === $mode || false === \in_array($mode[0], ['r', 'w', 'a', 'x', 'c'], true)) {
+                throw new \InvalidArgumentException('The mode ' . $mode . ' is invalid.');
+            }
+
+            throw new \RuntimeException('The file ' . $filename . ' cannot be opened.');
+        }
+
+        return self::create($resource);
     }
 
     /**
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -61,7 +61,6 @@ class Stream implements StreamInterface
      * @param string|resource|StreamInterface $body
      *
      * @throws \InvalidArgumentException
-     * @throws \ValueError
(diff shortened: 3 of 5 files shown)


src/Stream.php
```diff
@@ -61,6 +61,7 @@ private function __construct()
      * @param string|resource|StreamInterface $body
      *
      * @throws \InvalidArgumentException
+     * @throws \ValueError
```
GrahamCampbell, line 64: We should have a consistent API across all PHP versions. The internals of fopen changing should not effect the public API of this class.

src/UploadedFile.php
```diff
@@ -114,7 +114,13 @@ public function getStream(): StreamInterface
             return $this->stream;
         }
 
-        $resource = \fopen($this->file, 'r');
+        try {
+            $resource = \fopen($this->file, 'r');
+        } catch (\RuntimeException $e) {
```
GrahamCampbell, line 119: The fopen function doesn't ever throw this class of exception.

GrahamCampbell (changes requested):

GrahamCampbell:
I think it would be worth just copying Guzzle's tryFopen function: https://github.com/guzzle/psr7/blob/d7fe0a0eabc266c3dcf2f20aa12121044ff196a4/src/Utils.php#L343-L389.

composer.json
```diff
@@ -23,7 +23,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7.5 || 8.5 || 9.4",
         "php-http/psr7-integration-tests": "^1.0",
-        "http-interop/http-factory-tests": "^0.8",
+        "http-interop/http-factory-tests": "^0.7",
```
Nyholm, line 26: Let's use 0.9. (Just released)

Zegnat:
I am still divided on taking the Guzzle approach wholesale. At that point I would rather people just go and use the Guzzle implementation. That is the strength of interoperability interfaces, after all. I don’t think we should be adding utility classes/methods.

Current middle of the road solution: creating a new static factory method on our Stream implementation that handles it. This also lets us minimise some code in other places and centralises us to only a single file that uses `\fopen`. I have also marked this method as `@internal` to steer end-users towards the actual factories.

src/Stream.php
Nyholm: This can be removed, right?

src/Stream.php
```diff
@@ -88,6 +89,29 @@ public static function create($body = ''): StreamInterface
         throw new \InvalidArgumentException('First argument to Stream::create() must be a string, resource or StreamInterface.');
     }
 
+    /**
+     * Creates a new PSR-7 stream from path.
+     *
+     * @internal Implements Psr\Http\Message\StreamFactoryInterface::createStreamFromFile for internal use, library users should default to the actual Factory object
+     */
+    public static function createFromFile(string $filename, string $mode = 'r'): StreamInterface
```
Nyholm, line 97: If we kept this code in the factory and only surrounded the other two `fopen` with `try/catch`, wouldn't achieve the same thing and allow us to not have this method?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## Unreleased
```
Nyholm, line 5: Let's release it after merge =)

```suggestion
## 1.4.1
```

src/Stream.php
```diff
@@ -89,6 +89,29 @@ public static function create($body = ''): StreamInterface
...
+    {
+        try {
+            $resource = @\fopen($filename, $mode);
+        } catch (\Throwable $e) {
+            throw new \RuntimeException('The file ' . $filename . ' cannot be opened.');
+        }
+        if (false === $resource) {
+            if ('' === $mode || false === \in_array($mode[0], ['r', 'w', 'a', 'x', 'c'])) {
```
GrahamCampbell, line 105: strict in_array plz

GrahamCampbell (changes requested):

src/Stream.php
Zegnat: This same method is used by `moveTo` and `getStream` from `FileUpload`. Because they too create a `Stream` instance from a path.

src/Stream.php
Zegnat: Done. Is this faster internally? Because this check is always comparing between same types by virtue of its arguments.

src/Stream.php
Zegnat: I guess we can actually make it so there are no extra `fopen`s left at all with something like:

```diff
@@ -69,12 +69,6 @@ class Stream implements StreamInterface
             return $body;
         }
 
-        if (\is_string($body)) {
-            $resource = \fopen('php://temp', 'rw+');
-            \fwrite($resource, $body);
-            $body = $resource;
-        }
-
         if (\is_resource($body)) {
             $new = new self();
             $new->stream = $body;
@@ -86,6 +80,13 @@ class Stream implements StreamInterface
             return $new;
         }
 
+        if (\is_string($body)) {
+            $stream = self::createFromFile('php://temp', 'rw+');
+            $stream->write($body);
+
+            return $stream;
+        }
+
         throw new \InvalidArgumentException('First argument to Stream::create() must be a string, resource or StreamInterface.');
     }
 ```

nickdnk:
Hello guys

Is something currently holding this PR back? This package is now the only one that prevents me from running/testing my app on PHP 8.

nickdnk:
So it turns out it was psr7-server and not this one that held me back. I updated that to 1.0.x and everything is good (semantic beta-versioning pulling pranks on me). However, 1.4.1 from this package is not on Packagist, so I was just wondering if you forgot to tag it? It's on the changelog: https://github.com/Nyholm/psr7/blob/master/CHANGELOG.md - but not here: https://github.com/Nyholm/psr7/releases

GrahamCampbell:
Not forgotten. Intentionally not released yet. Pending https://github.com/Nyholm/psr7/pull/179 and possibly some other fixes before the release.

PR #177, opened 2021-04-22
creating a stream from string has unexpected behaviour

dbu:
when creating a stream from a string, the resource should be rewinded.


tests/StreamTest.php
```diff
@@ -44,6 +44,13 @@ public function testConvertsToString()
         $stream->close();
     }
 
+    public function testBuildFromString()
+    {
+        $stream = Stream::create('data');
+        $this->assertEquals('data', $stream->getContents());
```
dbu, line 50: RE the discussion in #176, i find it highly unexpected that this does not work.

dbu:
thanks!

Zegnat:
@Nyholm with this merged, should the next release be a major version bump per semver? Or do you want to run this as a bug fix only? It has the potential to break people’s implementations if they were building on top of what was previously a stable API from this library.

Nyholm:
I think the same principal should apply. 

You should never assume where the read pointer is. That is not part of any BC promise. 

> previously a stable API from this library.

That is true, but I am arguing that this never was a part of the API. The correct way is to always place your pointer after you read/write from a stream. 

But we could maybe be more explicit about this by releasing a minor version instead and with a clear note in the changelog. What do you think?

Nyholm:
> My argument would be that Stream::create() is part of the nyholm/psr7 API, even if not defined by PSR

True!

> The output of Stream::create('Hello') can no longer be chained with ->write(' world!') without the result being different. This to me means we have broken the BC promise with this PR.

Hm... Im not 100% convinced my point is valid. But we never specified this behaviour and it was untested.. 

I do think we should follow semver and I do think we should avoid doing a new major release. 

Currently we annoy the people that expect the stream to be rewinded, if we release this patch, we will annoy all other people... Hm..

Issue #181, opened 2021-05-27
getQueryParams() returns empty array by default, regardless of the URI


someniatko:
I've discovered this behavior (not to say "bug" yet) when tried ServerRequestInterface validation against the OpenAPI schema via the `league/openapi-psr7-validator` library. An OpenAPI schema may require a query parameter to be present in the request. And the mentioned library validates the presence of the query parameter by using the `getQueryParams()` method of the ServerRequestInterface, proof: https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/PSR7/Validators/QueryArgumentsValidator.php#L71-L72

And I support the choice of that library authors in that it's natural to expect the output of the `getQueryParams()` method to be populated with the query parameters from the URI. Although the interface allows to override them later via `withQueryParams(...)`, so that the method output, and the actual URI query may desync eventually, I still think it should be pre-populated with the params from the URI initially anyway.

If you feel that I, and the `league` team are wrong, and nobody should expect this method to be reliable, I'll instead open an issue on library's side, linking this one to it.

Zegnat:
Ah, yes, if you are not actually using `Nyholm/psr7-server` then maybe my response does not make a whole lot of sense! I have now moved the issue.

> Formally you are right that the PSR does not require these params to be pre-populated, but it's a reasonable expectation IMO.

Maybe. However, to keep the portability of PSR-7 implementations, I would say the only expectation that should be made from the PSR-7 spec design is that `getQueryParams()` will give you what was set by `withQueryParams()`.

In fact, I am not sure if `slim/psr7` does what I would expect it to do. Take this example:

```php
$uriFactory = new \Slim\Psr7\Factory\UriFactory();
$requestFactory = new \Slim\Psr7\Factory\ServerRequestFactory(
    new \Slim\Psr7\Factory\StreamFactory(),
    $uriFactory
);

$request1 = $requestFactory->createServerRequest('GET', 'https://example.com/?query=value');
var_dump($request1->getQueryParams());

$request2 = $requestFactory->createServerRequest('GET', 'https://example.com/?query=value');
$request2 = $request2->withUri($uriFactory->createUri('https://example.com/?after=creation'));
var_dump($request2->getQueryParams());
$request2 = $request2->withUri($uriFactory->createUri('https://example.com/?after=getQuery'));
var_dump($request2->getQueryParams());
```
I has the following output:
```
array(1) {
  ["query"]=>
  string(5) "value"
}
array(1) {
  ["after"]=>
[...]

PR #182, opened 2021-06-19
Revert #177

### LINKED ISSUE #180: Before releasing 1.5.0
Before we release the next version, I would like to discuss one thing. The discussion started [here](https://github.com/Nyholm/psr7/pull/177) between @Zegnat, @dbu and myself. A short summery is: 

```php
$s = \Nyholm\Psr7\Stream::create('Hello');
$s->write(' W');
echo "getContents(): \n";
echo $s->getContents();
echo "\n\n";
echo "toString(): \n";
echo $s->__toString();
```

In 1.4.0: 

```text
getContents(): 


toString(): 
Hello W

```

In master: 

```text
getContents(): 
llo

toString(): 
 Wllo

```

--------

If we dont use `->write()`

```php
$s = \Nyholm\Psr7\Stream::create('Hello');

echo "getContents(): \n";
echo $s->getContents();
echo "\n\n";
echo "toString(): \n";
echo $s->__toString();
```

In 1.4.0 (last release): 

```text
getContents(): 


toString(): 
Hello

```

In master: 

```text
getContents(): 
Hello

toString(): 
Hello

```

---------

Yes the PSR-7 `getContents()` method is confusing and should never be used. But the issue Im most concerned with is if this is a BC break or not. We have never specified where the read pointer should be after you run `Stream::create()` and we have **always** said that you never should assume anything. Ie, when you get a stream, you place the pointer yourself so you know where it is. 

This functionality has also never been tested (until now). The reason for making the change is because we have got so many confused users and complaints about this... so we are rather pragmatic than 100% strict/correct.

Nyholm:
This will revert #177 so we don't introduce new behaviour. This will fix #180. 

FYI @Zegnat @dbu. The discussion in #180 stalled a bit. By reverting this we are predictable (but maybe wrong), we are also consistent between creating a stream from string compared to from resource.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- `Stream::create` with a string needs to rewind the created memory stream.
 - `Psr17Factory::createStreamFromFile`, `UploadedFile::moveTo`, and
   `UploadedFile::getStream` no longer throw `ValueError` in PHP 8.
 
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -71,7 +71,6 @@ class Stream implements StreamInterface
         if (\is_string($body)) {
             $resource = \fopen('php://temp', 'rw+');
             \fwrite($resource, $body);
-            \rewind($resource);
             $body = $resource;
         }
 
--- a/tests/StreamTest.php
+++ b/tests/StreamTest.php
@@ -47,7 +47,8 @@ class StreamTest extends TestCase
     public function testBuildFromString()
     {
         $stream = Stream::create('data');
-        $this->assertEquals('data', $stream->getContents());
+        $this->assertEquals('', $stream->getContents());
+        $this->assertEquals('data', $stream->__toString());
         $stream->close();
     }
 


dbu (commented):
imho this is overthinking it.

if you really think its better to have it this way, i suggest to explain why the behaviour is this way in the StreamTest::testBuildFromString.

and to clarify the behaviour in the documentation.

the readme example on $psr17Factory->createStream should have the rewind and a note why you need to rewind.

and the phpdoc for Stream::create and for the psr17 factory createStream methods should also have a warning that you need to rewind the stream you just created if you want it to work.

Nyholm:
I agree with you that your way would make more sense. 

In #180 we discussed that this change would be a BC break because there might be application that will break if this was released. 

What is your opinion on the BC issue?

dbu:
> What is your opinion on the BC issue?

thats what i meant with "overthinking" it. i consider rewinding the stream a DX fix that can have side effects if somebody relied on undocumented behaviour. but i can't really claim it is a bug fix either, as the behaviour is unspecified in the spec, so it is not strictly a bug.

if it was a straight up bug, i think BC considerations would not make sense, but with this situation it is really an edge case. maybe we should simply release a new major if there is too much worry about side effects?

dbu:
in the end its your call. if i was the author of the library, i would consider it a fix and accept the risk that it could be a problem in an edge case... but it your library, so do what you think is right ;-)

Zegnat:
Like I said in the previous discussion: I do consider this a semver BC break, but have no real opinion on how strictly we are supposed to be following semver. I am very much of the opinion that not documenting the stream cursor position has been a big miss on the PSR mapping of PHP streams into PSR-7/PSR-17, and maybe when I have the time to do so I will try to reignite the discussion on that in the mailing list again.

I wonder if reverting this and then planning for a 2.0 release might be a good plan, where we specifically make the goal of 2.0 to compare with default behaviours of other PSR-7 implementations? #181 comes to mind as another one where the spec may be a slightly ambiguous and following convention might be a good idea. I would just like to see the data that there actually exists consensus amongst implementations :)

Zegnat (approved):

Nyholm:
I really do want to avoid a version 2.0 of this library.

Could we make sure the (soon to be released) Guzzlehttp/psr7:2.0 is aligned with us instead?

> I do consider this a semver BC break, but have no real opinion on how strictly we are supposed to be following semver.

Has this "feature" ever been documented? It sure has not been tested. If it has not been tested or documented, is it really a feature?

Zegnat:
> Has this "feature" ever been documented? It sure has not been tested.

Surely the fact that the PR changes a test, means it is tested? 😉

But yeah, as said in the original discussion, hard to say. Honestly this library does not make any other promises than PSR does. And with that said, PSR makes no promises about the cursor ever (as far as I can recall right now). So in some way, all behaviour around the cursor is undocumented behaviour. And the only way to ensure Stream is ever right is to rewind after the fact, to use `__toString()` to get all contents, etc.

Annoylingly enough, projects following pure PSR defined behaviour would never notice this difference. Because they should be using `__toString()` and manually managing cursor. It is only projects depending on the undocumented behaviour of PSR-7 that notice the difference between this library and others...

> Could we make sure the (soon to be released) Guzzlehttp/psr7:2.0 is aligned with us instead?

Have they answered the cursor question? Has it been documented why certain decisions were reached?

I would describe our behaviour to be:

* Creating a Stream from a string is an alias for `fopen` on a block of memory, writing to it, and then using that resource as stream.

To me it sounds like the alternative is:

* Creating a Stream from a string is an alias for creating a file with the string in it, `fopen` the file, and then using that resource as stream.

It is clear from PSR-7 that Steam object matches closely to the functions around `fopen`. I do not know what the most obvious way is to get the underlying stream resource when one needs to be created from a string. I personally think our way is very intuitive.

Nyholm:
We may be overthinking. But not merging this PR might lead to more head aches. 

I'll marge this and then tell users to never use `Stream::getContents()`.

PR #185, opened 2021-07-15
Improve error handling

nicolas-grekas:
The current logic breaks when eg `stream_get_contents()` triggers a warning after a timeout.
See https://github.com/symfony/symfony/issues/42100 for background.
guzzle/psr7 has the same issue btw.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -80,7 +80,7 @@ class UploadedFile implements UploadedFileInterface
 
         if (\UPLOAD_ERR_OK === $this->error) {
             // Depending on the value set file or stream variable.
-            if (\is_string($streamOrFile)) {
+            if (\is_string($streamOrFile) && '' !== $streamOrFile) {
                 $this->file = $streamOrFile;
             } elseif (\is_resource($streamOrFile)) {
                 $this->stream = Stream::create($streamOrFile);
@@ -114,11 +114,11 @@ class UploadedFile implements UploadedFileInterface
             return $this->stream;
         }
 
-        try {
-            return Stream::create(\fopen($this->file, 'r'));
-        } catch (\Throwable $e) {
-            throw new \RuntimeException(\sprintf('The file "%s" cannot be opened.', $this->file));
+        if (false === $resource = @\fopen($this->file, 'r')) {
+            throw new \RuntimeException(\sprintf('The file "%s" cannot be opened: %s', $this->file, \error_get_last()['message'] ?? ''));
         }
+
+        return Stream::create($resource);
     }
 
     public function moveTo($targetPath): void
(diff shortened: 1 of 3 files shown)


src/UploadedFile.php
```diff
@@ -114,11 +114,11 @@ public function getStream(): StreamInterface
             return $this->stream;
         }
 
-        try {
-            return Stream::create(\fopen($this->file, 'r'));
-        } catch (\Throwable $e) {
-            throw new \RuntimeException(\sprintf('The file "%s" cannot be opened.', $this->file));
+        if (false === $h = @\fopen($this->file, 'r')) {
+            throw new \RuntimeException(\sprintf('The file "%s" cannot be opened: %s', $targetPath, error_get_last()['message'] ?? ''));
```
stof, line 118: this must be `$this->file`

simPod:
```suggestion
            throw new \RuntimeException(\sprintf('The file "%s" cannot be opened: %s', $this->file, error_get_last()['message'] ?? ''));
```

nicolas-grekas: fixed thanks

src/UploadedFile.php
```diff
@@ -130,20 +130,23 @@ public function moveTo($targetPath): void
...
             if ($stream->isSeekable()) {
                 $stream->rewind();
             }
 
-            try {
-                // Copy the contents of a stream into another stream until end-of-file.
-                $dest = Stream::create(\fopen($targetPath, 'w'));
-            } catch (\Throwable $e) {
-                throw new \RuntimeException(\sprintf('The file "%s" cannot be opened.', $targetPath));
+            if (false === $h = @\fopen($targetPath, 'w')) {
```
Nyholm, line 144: $h?

Is this your C background that is showing itself?

src/UploadedFile.php
```diff
@@ -80,7 +80,7 @@ public function __construct($streamOrFile, $size, $errorStatus, $clientFilename
 
         if (\UPLOAD_ERR_OK === $this->error) {
             // Depending on the value set file or stream variable.
-            if (\is_string($streamOrFile)) {
+            if (\is_string($streamOrFile) && '' !== $this->file) {
```
Nyholm, line 83: I think `$this->file` will always be `null` here, right?

Nyholm (commented):
Thank you. I just have one question.

src/UploadedFile.php
nicolas-grekas: renamed :)

src/UploadedFile.php
nicolas-grekas: right, fixed!

Nyholm (approved):
Thank you

simPod:
@Nyholm would you mind releasing a new patch? Thanks!

PR #186, opened 2021-08-09
Stupid error

jorgdz:
* Fix the method createStreamFromFile animals >:(


Nyholm:
Could you elaborate why this is needed? 

According to the [PSR-17](https://www.php-fig.org/psr/psr-17/) specification, we do need to open the file.

frikiluser:
> Could you elaborate why this is needed?
> 
> According to the [PSR-17](https://www.php-fig.org/psr/psr-17/) specification, we do need to open the file.

Probably the $mode should be checked/sanitized before opening the file. But I think it's not the point of this issue.

PR #187, opened 2021-08-10
Add explicit `@return mixed` phpdoc

nicolas-grekas:
This will allow removing one deprecation notice from the Symfony codebase:

> Method "Psr\Http\Message\StreamInterface::getMetadata()" will return "mixed" as of its next major version. Doing the same in implementation "Nyholm\Psr7\Stream" will be required when upgrading.


Nyholm (approved):
Thank you

PR #188, opened 2021-11-13
Add return types to HttplugFactory

ostrolucky:
This is needed to do now because symfony/error-handler triggers deprecation notices such these

```
  1x: Method "Http\Message\RequestFactory::createRequest()" might add "RequestInterface" as a native return type declaration in the future. Do the same in implementation "Nyholm\Psr7\Factory\HttplugFactory" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in DiscoveredClientsTest::testForcedDiscovery from Http\HttplugBundle\Tests\Functional

  1x: Method "Http\Message\ResponseFactory::createResponse()" might add "ResponseInterface" as a native return type declaration in the future. Do the same in implementation "Nyholm\Psr7\Factory\HttplugFactory" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in DiscoveredClientsTest::testForcedDiscovery from Http\HttplugBundle\Tests\Functional

  1x: Method "Http\Message\StreamFactory::createStream()" might add "StreamInterface" as a native return type declaration in the future. Do the same in implementation "Nyholm\Psr7\Factory\HttplugFactory" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in DiscoveredClientsTest::testForcedDiscovery from Http\HttplugBundle\Tests\Functional
```

### DIFF
--- a/src/Factory/HttplugFactory.php
+++ b/src/Factory/HttplugFactory.php
@@ -6,6 +6,9 @@ namespace Nyholm\Psr7\Factory;
 
 use Http\Message\{MessageFactory, StreamFactory, UriFactory};
 use Nyholm\Psr7\{Request, Response, Stream, Uri};
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UriInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


Nyholm (approved):
I am happy with this. 
It is technically a BC break, but the class is marked as final. 

FYI, We could also implement this by adding: 

```php
/**
 * @return RequestInterface
 */
 #[\ReturnTypeWillChange]
 ```

Nyholm:
Thank you for this contribution.

acirtautas:
Hi @Nyholm and @ostrolucky, thank you for your great work!
As the change has been merged it would be good to tag an release it.

Nyholm:
Yes, tagging now

PR #202, opened 2022-08-11
Fixed rawurlencode for userInfo in Uri

ezimuel:
When you specify a username and password using `Uri::withUserInfo()` these are not encoded for URL (see [here](https://github.com/elastic/elastic-transport-php/pull/7#issuecomment-1171086850)). For instance, if the username is `foo@` and the password is `bar` the userInfo results in `foo@:bar` that is not a valid URL encoded string.

Other PSR-7 implementations use `rawurlencode()` for encoding the characters in the URL (e.g. [Guzzle](https://github.com/guzzle/psr7/issues/252), [Laminas-diactoros](https://github.com/laminas/laminas-diactoros/blob/283204ee45bec928f86229deaf8ca54ec0bc8d33/src/Uri.php#L538-L549)).

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -145,12 +145,12 @@ class Uri implements UriInterface
     public function withUserInfo($user, $password = null): self
     {
         if ($user !== \rawurlencode($user)) {
-            throw new \InvalidArgumentException("The user contains invalid URL characters. Please use rawurlencode()");
+            throw new \InvalidArgumentException('The user contains invalid URL characters. Please use rawurlencode()');
         }
         $info = $user;
         if (null !== $password && '' !== $password) {
             if ($password !== \rawurlencode($password)) {
-                throw new \InvalidArgumentException("The password contains invalid URL characters. Please use rawurlencode()");
+                throw new \InvalidArgumentException('The password contains invalid URL characters. Please use rawurlencode()');
             }
             $info .= ':' . $password;
         }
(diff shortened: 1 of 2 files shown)


ezimuel:
@Nyholm can you review this PR? Thanks.

Nyholm:
Merging this PR will break all existing usage of this feature. 
I just quickly checked PSR7 specification and it does not mention anything about URL encoding. 

I think it would be a better solution to throw an exception if you try to add UserInfo with characters not allowed.

ezimuel:
@Nyholm I see your point. I did the change throwing an `InvalidArgumentException` if the user or the password is not URL encoded.

Nyholm:
Thank you. 

Let's refer to [rfc3986](https://www.rfc-editor.org/rfc/rfc3986#page-18) to determine if what is valid or not. 
I think it is this. Please verify: 

```
user:
  REGEX: [a-zA-Z0-9-\._~%!\$&'\(\}\*\+,;=]+

password:
  REGEX: [a-zA-Z0-9-\._~%!\$&'\(\}\*\+,;=]+
```

We also need tests for this.

ezimuel:
@Nyholm  I checked [rfc3986](https://www.rfc-editor.org/rfc/rfc3986#page-18) and the correct regex is as follows:
```
^(?:[a-zA-Z0-9_\-\.~!\$&\'\(\)\*\+,;=:]|%[A-Fa-f0-9]{2})+$
```
We need to check for valid `% HEXDIG HEXDIG` that's why I used `%[A-Fa-f0-9]{2}`.
I provided also some cases for unit testing.

ezimuel:
@Nyholm I was thinking that we should throw two different exceptions for the user and for the password. Can we add specific Exception classes extending `\InvalidArgumentException`?

nicolas-grekas:
Closing in favor of #213

PR #203, opened 2022-10-10
Improve extensibility by removing `@final` and making Stream's constructor public

nicolas-grekas:
I'm working on a situation where I'd like to have control over the signature of the constructor of PSR-7 implementations.
I managed to use inheritance for all implementations but nyholm/psr7
That's why I'm proposing this change. It improves extensibility without hurting maintainability IMHO.


Nyholm (commented):
Im okey with adding a constructor. But Im hesitant to remove `@final`.

nicolas-grekas:
IMHO, the `@final` annotations are the one remaining thing that might prevent wider adoption of this lib. The reason is that there *are* legitimate use cases for extending the classes (eg mine ;))

I think we need to figure out *why* we're using these annotations. If the reason is because they make maintenance easier, then that's legit. But if the reason is to "annoy" people because we think they should not extend the classes, then I want us to reconsider this reasoning, because then the reason becomes arbitrary, as in "us, the authors, think we know better than you, and we restrict you from doing that, because we decided so."

So, does `@final` help maintenance in any way? I don't think so.

PR #205, opened 2022-11-16
fix psr17factory create stream

### LINKED ISSUE #204: Empty body when using psr17factory's createRequest
Using the psr17factory for creating a request object using this syntax, when we try to get the contents of the body, we get empty string.
```
$factory = new Psr17Factory();
$req = $factory->createRequest('POST', 'https://nyholm.tech');
$body = $factory->createStream('Foo');
$req = $req->withBody($body);
var_dump($req->getBody()->getContents()); // prints empty string.
```
The problem is in the createStream method, while creating the stream the file seek goes to the end of the temp file created and when we try to get the contents it starts reading from end, hence the empty string.

nawedr:
Closes #204 
As a fix, when we have created the stream, we can set the seek again to start of the file so that when we invoke  `$req->getBody()->getContents();` we get the actual body and not empty string.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -71,6 +71,7 @@ class Stream implements StreamInterface
         if (\is_string($body)) {
             $resource = \fopen('php://temp', 'rw+');
             \fwrite($resource, $body);
+            \fseek($resource, 0);
             $body = $resource;
         }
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
AFAIK, this behavior is not specified by PSR-7 so the current behavior is also correct.
What could help is add a test case for this in https://github.com/php-http/psr7-integration-tests/ and ensuring all implementations agree on the expected behavior first.

Meanwhile, I'm closing here, please let us know.

PR #207, opened 2023-01-03
make response serializable

VirgileZol:


nicolas-grekas:
This is not explicitly allowed by PSR-7, so doing that would enable building on a non-specified behavior, enabling tight-coupling to one specific implementation.

Instead, I'd advise not serializing PSR-7 objects directly.

An alternative course of action might be to add a test case to https://github.com/php-http/psr7-integration-tests/ and ensure that all major implementations agree on this behavior. That'd create a sort of de facto standard.

Please let us know if this happens.
Closing meanwhile.

PR #213, opened 2023-03-30
Encode userinfo (but don't double-encode)

### LINKED ISSUE #206: Filter Uri->withUserInfo() input
I have a login, which contains `@` in username and password. User info with this character is currently not encoded. I think there should be a `filterUserInfo()` method to correctly encode input as other with* methods do.

I checked other popular implementations and all of them do that:
- https://github.com/guzzle/psr7/blob/67c26b443f348a51926030c83481b85718457d3d/src/Uri.php#L436-L439
- https://github.com/slimphp/Slim-Psr7/blob/7e7e63e3b624ed07bde14fbd9f91ab8904402375/src/Uri.php#L174-L177
- https://github.com/laminas/laminas-diactoros/blob/6028af6c3b5ced4d063a680d2483cce67578b902/src/Uri.php#L275-L278

Most accurate seem to be guzzle and laminas implementations which do this:

```php
    private function filterUserInfo($userInfo): string
    {
        if (!is_string($userInfo)) {
            throw new \InvalidArgumentException('User info must be a string');
        }

        return preg_replace_callback(
            '/(?:[^%' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . ']+|%(?![A-Fa-f0-9]{2}))/',
            [$this, 'rawurlencodeMatchZero'],
            $userInfo
        );
    }
```

Only difference is in laminas/laminas-diactoros which also checks for invalid UTF-8 characters

nicolas-grekas:
Fix #206

Yes, this is unspecified by the PSR, but the current behavior produces broken URLs, and other mainstream implementations do the same: encoding characters that must be, while preventing double-encoding. I think this makes sense. It's more useful than an exception.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "symfony/error-handler": "^4.4"
     },
     "provide": {
+        "php-http/message-factory-implementation": "1.0",
         "psr/http-message-implementation": "1.0",
         "psr/http-factory-implementation": "1.0"
     },
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -144,9 +144,9 @@ class Uri implements UriInterface
 
     public function withUserInfo($user, $password = null): self
     {
-        $info = $user;
+        $info = \rawurlencode(\rawurldecode($user));
         if (null !== $password && '' !== $password) {
-            $info .= ':' . $password;
+            $info .= ':' . \rawurlencode(\rawurldecode($password));
         }
 
         if ($this->userInfo === $info) {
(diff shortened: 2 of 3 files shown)


dbu:
if i understand correctly, PSR7 is a bit underspecified in that regard. would it be possible to amend PSR-7 to make it the official behaviour? maybe with a SHOULD for the current version and a MUST for a new major version?

nicolas-grekas:
That could be a path forward for PSR-7 yes @dbu. For now, I'm advising ppl to contribute these edge cases to https://github.com/php-http/psr7-integration-tests/. Like with https://github.com/php-cache/integration-tests/, this would give a reference test suite to validate implementations, and spot+define behaviors for unspecified cases.

PR #221, opened 2023-03-31
Encode only reserved characters in user-info

### LINKED ISSUE #219: Escaping of user info breaks NTLM authentication
The change in #213 breaks ntml based authentication when specifying a domain.

When using ntml auth it's common to have a domain that is part of the username like this
```
curl_setopt($ch, CURLOPT_USERPWD, 'DOMAIN\username:password');
```

The slash is not suppose to be escaped and if it's the authentication fails

nicolas-grekas:
Fix #219

This is enough to produce non-broken URLs.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -23,6 +23,8 @@ class Uri implements UriInterface
 
     private const CHAR_SUB_DELIMS = '!\$&\'\(\)\*\+,;=';
 
+    private const CHAR_GEN_DELIMS = ':\/\?#\[\]@';
+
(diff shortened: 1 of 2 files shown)


dbu (commented):
seems all the implementations do it this way, good fix.

dbu:
btw, you might want to mention this in the changelog. there is an edge case when people use getUserInfo and assume that the info is *not* urlencoded it will now be urlencoded...

nicolas-grekas:
See #218 for the changelog proposal.

PR #223, opened 2023-04-05
chore(deps): allow psr/http-message v2

simPod:


nicolas-grekas:
In the CI:

>   - Locking psr/http-message (1.1)

So that's not enough.

Also, is this lib really compatible with v2? Doesn't it miss some return types?

simPod:
Hm, let's test it.

simPod:
There are indeed some type changes required but I believe the v2 support won't be possible now since old php versions are allowed.

src/MessageTrait.php
```diff
@@ -81,17 +77,17 @@ public function getHeaderLine($header): string
         return \implode(', ', $this->getHeader($header));
     }
 
-    public function withHeader($header, $value): self
+    public function withHeader(string $name, $value): self
```
nicolas-grekas, line 80: the name should not be changed

src/Response.php
```diff
@@ -65,7 +65,7 @@ public function getReasonPhrase(): string
         return $this->reasonPhrase;
     }
 
-    public function withStatus($code, $reasonPhrase = ''): self
+    public function withStatus(int $code, string $reasonPhrase = ''): ResponseInterface
```
nicolas-grekas, line 68: `self` should be kept everywhere

src/Response.php
```diff
@@ -65,7 +65,7 @@ public function getReasonPhrase(): string
...
     {
         if (!\is_int($code) && !\is_string($code)) {
```
nicolas-grekas, line 70: can be removed

src/ServerRequest.php
```diff
@@ -79,7 +79,7 @@ public function getUploadedFiles(): array
     /**
      * @return static
      */
-    public function withUploadedFiles(array $uploadedFiles)
+    public function withUploadedFiles(array $uploadedFiles): ServerRequestInterface
```
nicolas-grekas, line 82: we should use `self` everywhere instead

src/ServerRequest.php
```diff
@@ -150,7 +150,7 @@ public function getAttributes(): array
     /**
      * @return mixed
      */
-    public function getAttribute($attribute, $default = null)
+    public function getAttribute(string $attribute, $default = null)
     {
         if (!\is_string($attribute)) {
```
nicolas-grekas, line 155: drop

src/ServerRequest.php
```diff
@@ -163,7 +163,7 @@ public function getAttribute($attribute, $default = null)
         return $this->attributes[$attribute];
     }
 
-    public function withAttribute($attribute, $value): self
+    public function withAttribute(string $attribute, $value): ServerRequestInterface
     {
         if (!\is_string($attribute)) {
```
nicolas-grekas, line 168: drop

src/ServerRequest.php
```diff
@@ -175,7 +175,7 @@ public function withAttribute($attribute, $value): self
         return $new;
     }
 
-    public function withoutAttribute($attribute): self
+    public function withoutAttribute(string $attribute): ServerRequestInterface
     {
         if (!\is_string($attribute)) {
```
nicolas-grekas, line 180: drop

src/Stream.php
```diff
@@ -293,7 +290,7 @@ public function getContents(): string
     /**
      * @return mixed
      */
-    public function getMetadata($key = null)
+    public function getMetadata(string $key = null)
     {
         if (null !== $key && !\is_string($key)) {
```
nicolas-grekas, line 295: drop

src/Uri.php
```diff
@@ -138,7 +138,7 @@ public function getFragment(): string
         return $this->fragment;
     }
 
-    public function withScheme($scheme): self
+    public function withScheme(string $scheme): UriInterface
     {
         if (!\is_string($scheme)) {
```
nicolas-grekas, line 143: drop (I'm stopping here, please continue :)

src/Uri.php
```diff
@@ -155,7 +155,7 @@ public function withScheme($scheme): self
         return $new;
     }
 
-    public function withUserInfo($user, $password = null): self
+    public function withUserInfo(string $user, ?string $password = null): UriInterface
```
nicolas-grekas, line 158: self everywhere, please continue :)

nicolas-grekas:
We should bump to PHP >=7.2 also

simPod:
If that's okay with you I've opened `>=7.2 PR` here https://github.com/Nyholm/psr7/pull/226

simPod:
Also, I think we cannot have support for both v1 and v2 since methods with typed arguments are not backward compatible.

nicolas-grekas:
> methods with typed arguments are not backward compatible

because of the behavior change when passing incompatible types, or?

simPod:
Because changing e.g.

```diff
-     public function withProtocolVersion($version): self
+     public function withProtocolVersion(string $version): self
```

is required and compatible with v2 but not compatible with v1

nicolas-grekas:
It's compatible with 1.1, so we could just bump to ^1.1|^2.0, no?

src/MessageTrait.php
simPod: Ok but it has been changed in psr/http-message v2

nicolas-grekas: argument names are not part of the contract in PHP

simPod:
Test suite is being adapted here https://github.com/php-http/psr7-integration-tests/pull/68

nicolas-grekas:
Closing in favor of #234, thanks for pushing for this.

PR #229, opened 2023-04-09
Extract path normalization in a separate method

kudashevs:
Because the leading slashes path normalization seems to be duplicated, I would like to propose to extract it to a separate method.


kudashevs:
Wrong assumption, sorry. The functionality is different and cannot be extract.

PR #230, opened 2023-04-09
Use zval storage for Stream instances created from strings

nicolas-grekas:
Alternative to #227 so that we can compare both approaches.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -81,10 +81,7 @@ class Stream implements StreamInterface
         }
 
         if (\is_string($body)) {
-            $resource = \fopen('php://temp', 'rw+');
-            \fwrite($resource, $body);
-            \fseek($resource, 0);
-            $body = $resource;
+            $body = self::openZvalStream($body);
         }
 
         if (!\is_resource($body)) {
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
PR is ready, I confirm this prevents the memcopy from happening.

src/Stream.php
```diff
@@ -313,4 +310,104 @@ public function getMetadata($key = null)
 
         return $meta[$key] ?? null;
     }
+
+    private static function openZvalStream(string $body)
+    {
+        static $wrapper;
+
+        $wrapper ?? \stream_wrapper_register('Nyholm-Psr7-Zval', $wrapper = \get_class(new class() {
```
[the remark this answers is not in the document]
nicolas-grekas: It's not registered twice, where do you read that? About `??=`, PHP 7.1 rulez. For the "if", I find the current style more readable, because less indentation.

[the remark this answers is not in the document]
mvorisek: I would personally prefer to not do it as any unrelated code MUST never touch wrappers that are not owned by the code.

src/Stream.php
```diff
@@ -313,4 +310,104 @@ public function getMetadata($key = null)
...
+            }
+
+            public function stream_set_option(): bool
+            {
+                return true;
+            }
+
+            public function stream_truncate(int $new_size): bool
+            {
+                if ($new_size) {
```
mvorisek, line 392: the if condition and else should not be needed

nicolas-grekas: it's an small optimization

src/Stream.php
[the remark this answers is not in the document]
nicolas-grekas: I prefer keeping it, adding some guards against global scope manipulations.

src/Stream.php
mvorisek: I belive `substr(expr, 0, 0)` is optimized internally to return interned `''`.

nicolas-grekas: the optimization is about removing the need to call any function :)

nicolas-grekas:
PR rebased, still green. I updated the branch-alias also.

Nyholm (approved):
Thank you

PR #241, opened 2023-04-20
Improve performance of creating streams from strings

nicolas-grekas:
Using https://github.com/devanych/psr-http-benchmark, I figured out that for strings smaller than ~200k, `php://memory` is faster than the custom zval stream. Past this threshold, zval stream is way faster thanks to copy-on-write.


Nyholm (approved):
Thank you for testing

PR #247, opened 2023-06-14
Test CI

Nyholm:


stof:
@Nyholm this won't work. The issue is that you use git inside your docker image, which has a separate global git config. The fix needs to happen inside the docker image (or you could do like https://github.com/async-aws/aws and stop using the docker image)

PR #252, opened 2023-11-10
Fix error handling in Stream::getContents()

nicolas-grekas:
`stream_get_contents()` swallows errors. This creates issues like https://github.com/symfony/symfony/issues/52490, where errors reported by the stream layer aren't propagated to user-land.

This is even in the C source:
https://github.com/php/php-src/blob/311cae03e730c76aed343312319ed8cf1c37ade0/main/streams/streams.c#L1512

Using `fread()`, allows fixing this issue.


nicolas-grekas:
guzzle/psr7 looks OK BTW:
https://github.com/guzzle/psr7/blame/2.6/src/Utils.php#L404

nicolas-grekas:
laminas/diactoros is affected: https://github.com/laminas/laminas-diactoros/blob/3.4.x/src/Stream.php#L282 /cc @weierophinney

stof:
@nicolas-grekas is this a case for which we could write a test in `php-http/psr7-integration-tests` so that all PSR-7 implementations could know whether they handle that case properly by running the shared testsuite ? Or is it not testable this way ?

nicolas-grekas:
slim/psr7 is affected: https://github.com/slimphp/Slim-Psr7/blob/4299f1650acad6df77fbaeb6f18c59e46af761b9/src/Stream.php#L357 /cc @l0gicgate

nicolas-grekas:
ringcentral/psr7 affected too: https://github.com/ringcentral/psr7/blob/360faaec4b563958b673fb52bbe94e37f14bc686/src/Stream.php#L102 /cc @mtdowling

nicolas-grekas:
workerman/psr7 affected: https://github.com/walkor/psr7/blob/e21521b72b5a891bd1a7a51376da12801863ce5b/src/Stream.php#L99 /cc @walkor

nicolas-grekas:
httpsoft/http-message affected too: https://github.com/httpsoft/http-message/blob/6d86446b9d8f92cc06c60375d3e67cf3407a57a5/src/StreamTrait.php#L332 /cc @devanych

nicolas-grekas:
I added a test case at https://github.com/php-http/psr7-integration-tests/pull/74 and updated the implementation to cover more situations (like guzzle does).

vbartusevicius:
Hi @nicolas-grekas, thanks for fixing such a lurking bug! @Nyholm, if possible, when can we expect a release with the fix?

Nyholm:
It was released in 1.8.1. 

See https://github.com/Nyholm/psr7/releases/tag/1.8.1

src/Stream.php
```diff
@@ -260,11 +260,19 @@ public function getContents(): string
...
+        \set_error_handler(static function ($type, $message) use (&$exception) {
+            throw $exception = new \RuntimeException('Unable to read stream contents: ' . $message);
+        });
+
+        try {
+            return \stream_get_contents($this->stream);
+        } catch (\Throwable $e) {
+            throw $e === $exception ? $e : new \RuntimeException('Unable to read stream contents: ' . $e->getMessage(), 0, $e);
+        } finally {
+            \restore_error_handler();
```
shadowhand, line 274: I don't understand how setting an error handler fixes this issue when [other tools](https://github.com/thecodingmachine/safe/blob/3a6e43b4be3d3d65b272c2660300a50658ce168b/generated/stream.php#L221-L235) and [even Guzzle](https://github.com/guzzle/psr7/blob/2.6/src/Utils.php#L420-L422) check for a `false` return. Nothing about this code seems like it will actually solve the issue, because the return value of `stream_get_contents` is not being checked.

shadowhand:
It might also be helpful to have a test case for this in [http-factory-tests](https://github.com/http-interop/http-factory-tests/blob/main/test/StreamFactoryTestCase.php).

src/Stream.php
nicolas-grekas: When stream_get_contents returns false, it first always trigger a PHP warning. This we catch those in the error handler, the function will never really return false.

shadowhand: Considering that such behavior is [not documented](https://www.php.net/stream_get_contents), why take that approach over checking the return value is false? Every other fix I've looked at around this issue is doing a `$ret === false` check...

shadowhand: > That would be dead code.

Are you saying that checking `=== false` does not work, or are you saying that this monkey patching with the error handler identifies the _impending_ `return false` more quickly?

nicolas-grekas: I'm saying that the function will never return false because the error handler will always throw before.

shadowhand: But, if the error handler monkey patching was not there, we could rely on `$ret === false`, couldn't we?

In any case, I feel like either:

- this code needs a comment describing what it does, given that it relies on undocumented behavior
- it should rely on `$ret === false` checks, which the code did prior to this patch (but perhaps not correctly?)

shadowhand: Also, I want to be clear that I really appreciate you identifying this issue and submitting the PR. It only bothers me that we have MANY disparate projects with different solutions to the same problem... perhaps we should all be using [safe](https://github.com/thecodingmachine/safe) instead?

nicolas-grekas: If we don't set our own error handler, then we rely the existing, which can be a custom and broken one. But putting our own for the operation, we make things safe.

stof: See the link to the PHP code in the PR description. When doing a partial read, PHP does not return false right now because it does not propagate the `false` return value.

shadowhand: @stof 😕 and is there a related PHP bug report and/or PR?

nicolas-grekas: > perhaps we should all be using (and contributing to) [safe](https://github.com/thecodingmachine/safe) instead?

"safe" is too opinionated to be used in a generic package IMHO ;)

shadowhand: @nicolas-grekas the current "use this arcane code instead" situation being applied to a multitude of projects doesn't seem like a good solution either. Would adding `safe` really be such a terrible thing? And if safe is too opinionated, why not contribute your knowledge to making it something that package developers would support?

weierophinney: Considering `safe` is completely opt-in (you have to explicitly use the `Safe\\` namespace to use the functions it guards), it seems like it's up to the individual package maintainers as to whether or not it's "too opinionated".

odan: I have a question to understand this patch a little better. 
Why do you declare an "$exception" variable for the Exception although the Catch block gets the same Exception instance? Is there some "PHPish" specific logic / behavior behind this (that is not documented)?

nicolas-grekas: If we remove this check, we would always wrap the exception thrown in the error handler in another exception. This check allows removing this needless wrapping exception.

odan: Ok, thanks. Then my next question is, why not move the `set_error_handler` into the try-block block to handle that scope in the same place?

nicolas-grekas: if set_error_handler fails for any reasons, we don't want restore_error_handler to be called

stof: And in PHP, try/catch does not create a scope anyway.

devanych:
Hi @nicolas-grekas, thanks :+1:

